PR #17, opened 2021-04-26
docs: docs for some docless components

JoeDuncko:
## What is the problem?

Docs were missing for some components

- [x] Docs for useMountEffect
- [x] Docs for useRerender
- [x] Docs for useToggle
- [x] Docs for useUnmountEffect

## Checklist

  - _link issue here_

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,9 +57,9 @@ first.
    - Components representing hook functionality should be placed in file named after the hook
      with `.stories` suffix.  
      4ex: `useFirstMountState.stories.tsx`.
-   - Preferred format to write the docs is MDX.
-   - Add docs link and hook summary to the `README.md`.
-5. After all above steps are done - run `yarn lint:fix` and ensure that everything is styled by our standards.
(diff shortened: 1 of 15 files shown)


JoeDuncko:
Question: Why is the `effect` of `useMountEffect` of type `CallableFunction` instead of `EffectCallback`?

JoeDuncko:
@xobotyi Okay, I think I got them all. I'd definitely look over this as I'm not super familiar with Typescript. Feedback is appreciated!

xobotyi:
> Question: Why is the `effect` of `useMountEffect` of type `CallableFunction` instead of `EffectCallback`?
> 

Because `EffectCallback` can be void or destructor type, while `useMountEffect` expecting any callable function with any return type.
For the case of destruction required developers expected to use `useEffect`

JoeDuncko:
> > Question: Why is the `effect` of `useMountEffect` of type `CallableFunction` instead of `EffectCallback`?
> 
> Because `EffectCallback` can be void function or return destructor type, while `useMountEffect` expecting any callable function with any return type.
> For the case of destruction required developers expected to use `useEffect`.

Thanks!

PR #24, opened 2021-04-28
feat: new hook useIsMounted

xobotyi:
## What new hook does?

Returns function that yields current mount state.

## Checklist

  - _link issue here_

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ import { useMountEffect } from "@react-hookz/web/esnext";
     — Return boolean that is `true` only on first render.
   - [`useMountEffect`](https://react-hookz.github.io/web/?path=/docs/lifecycle-usemounteffect)
     — Run effect only when component is first mounted.
-  - [`useMountedGetter`](https://react-hookz.github.io/web/?path=/docs/lifecycle-usemountedgetter)
+  - [`useIsMounted`](https://react-hookz.github.io/web/?path=/docs/lifecycle-useismounted)
     — Returns function that yields current mount state.
   - [`useRerender`](https://react-hookz.github.io/web/?path=/docs/lifecycle-usererender)
     — Return callback that re-renders component.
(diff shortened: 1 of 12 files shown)


JoeDuncko:
It looks like this is supposed to be equivalent to react-use's ` useMountedState`. 

I'm not sure how I feel about this implementation. 
- Why not include the word `state` in the name? I think emulating something like `recoil`'s `useRecoilVallue` here... Maybe in that case this should be `useMountedValue`? Either way, not a big fan of using `getter` here - not very react-y IMO.
- Why does this return a function instead of a value?

xobotyi:
1. It is not a state, it is, as said, getter - hook reruns a function that yield component's mount state. I've named it incorrectly in `use-hooks`, people was confused sometimes. 
2. It is required mostly for async effects, as conventional state change unable to take effect over already triggered request callbacks. Thus we need kind of ref, but writing each time `isMounted.current` not handy.

xobotyi:
@JoeDuncko what i am thinking of is to rename it to `useIsMounted`

JoeDuncko:
1. understood
2. that makes sense

I definitely like `useIsMounted` more. Let's do that.

xobotyi:
@JoeDuncko otherwise fine? merge?

JoeDuncko:
Let me test the example, then I'll merge...

JoeDuncko:
```
Child HtmlWebpackCompiler:
                          Asset      Size               Chunks  Chunk Names
    __child-HtmlWebpackPlugin_0  6.23 KiB  HtmlWebpackPlugin_0  HtmlWebpackPlugin_0
    Entrypoint HtmlWebpackPlugin_0 = __child-HtmlWebpackPlugin_0
    [./node_modules/html-webpack-plugin/lib/loader.js!./node_modules/@storybook/core-common/dist/cjs/templates/index.ejs] 1.98 KiB {HtmlWebpackPlugin_0} [built]
ModuleNotFoundError: Module not found: Error: Can't resolve './useMountedGetter.stories' in '/Users/joeduncko/Documents/GitHub/web/stories'
```

xobotyi:
Aha, somehy not renamed during files renaming.

JoeDuncko (approved):
LGTM

PR #34, opened 2021-05-02
docs: alphabetize hooks README like storybook

JoeDuncko:
## What is the problem?

Hooks list in README didn't line up with storybook.

The hooks list now lines up with storybook.

## Checklist

  - _link issue here_


xobotyi:
Ive been thinking about it to alphabetize it somehow.
Merging for now, will come up with solution later.

PR #35, opened 2021-05-02
feat: new hook, useNetworkState

xobotyi:
## What new hook does?

Tracks the state of browser's network connection.

## Checklist

  - _link issue here_

### DIFF
--- a/src/useNetworkState.ts
+++ b/src/useNetworkState.ts
@@ -1,0 +1,137 @@
+import { useEffect } from 'react';
+import { isBrowser, noop } from './util/const';
+import { IInitialState } from './util/resolveHookState';
+import { useSafeState } from './useSafeState';
+import { off, on } from './util/misc';
+
+export interface INetworkInformation extends EventTarget {
+  readonly downlink: number;
+  readonly downlinkMax: number;
+  readonly effectiveType: 'slow-2g' | '2g' | '3g' | '4g';
+  readonly rtt: number;
+  readonly saveData: boolean;
+  readonly type:
+    | 'bluetooth'
+    | 'cellular'
+    | 'ethernet'
+    | 'none'
+    | 'wifi'
+    | 'wimax'
+    | 'other'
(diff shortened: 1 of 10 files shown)


src/useNetworkState.ts
```diff
@@ -0,0 +1,137 @@
...
+
+function getConnectionState(previousState?: IUseNetworkState): IUseNetworkState {
+  const online = navigator?.onLine;
+  const previousOnline = previousState?.online;
+
+  return {
+    online,
+    previous: previousOnline,
+    since: online !== previousOnline ? new Date() : previousState?.since,
+    downlink: conn?.downlink,
```
JoeDuncko, line 94: Why not expose all of `conn` here?

stories/useNetworkState.story.mdx
```diff
@@ -0,0 +1,49 @@
+import { Example } from './useNetworkState.stories';
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+
+<Meta title={'Sensor/useNetwork'} component={Example} />
+
+# useNetwork
+
+Tracks the state of browser's network connection.
+
+> As of the standard, it is not guaranteed that browser connected to the _Internet_, it only
```
JoeDuncko, line 10: Maybe we should quote MDN here and add a link?

`while you can assume that the browser is offline when it returns a false value, you cannot assume that a true value necessarily means that the browser can access the internet`

https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine

JoeDuncko (changes requested):

src/useNetworkState.ts
xobotyi: `navigator.connection` is a mutable object, also it is EventEmitter instance and it becomes kind of abstraction leak if to simply return it.

JoeDuncko: Got it

PR #42, opened 2021-05-06
feat: new hook `useSyncedRef`

xobotyi:
## What new hook does?

Like `useRef`, but it returns immutable ref that contains actual value.

I'm not sure about hook naming, i dont really like the `useSyncedRef`, but it is the best i came up with. 

## Checklist

  - _link issue here_

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,3 +90,8 @@ import { useMountEffect } from "@react-hookz/web/esnext";
 
     - [`useNetworkState`](http://react-hookz.github.io/?path=/docs/sensor-usenetwork--example)
       — Tracks the state of browser's network connection.
+
+- #### Miscellaneous
+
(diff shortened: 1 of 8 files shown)


JoeDuncko:
What's an example use case for this hook? It doesn't seem like the example story illustrates one very well.

xobotyi:
I'm unable to imagine different =(

JoeDuncko:
Do you have an example where this hook would be useful in general? I think I need help "getting" this one.

xobotyi:
It is mostly required for other hooks, to provide stable API, 4ex:
https://github.com/react-hookz/web/blob/master/src/useMediatedState.ts#L21

Will be replaced with this hook.

JoeDuncko (approved):
Got it, thanks for pointing that out! Let's roll with it

Issue #33, opened 2021-05-02
Port remaining hooks from `react-use`


JoeDuncko:
Our goal with `@react-hookz/web` is to give the react community a general-purpose React hooks library built with care.

We'd like to port the remaining hooks from [react-use](https://github.com/streamich/react-use) (the project `@react-hookz/web` grew out of) while keeping in mind our tenants:

- General-purposeness (hooks should fulfill a wide array of use cases)
- Composability (a preference for hooks that are useful for building other hooks)
- No or _very few_ dependencies
- SSR compatibility
- 100% test coverage (ideally)
- Concrete example use cases

### Sensors

- [ ] [useBattery](https://github.com/streamich/react-use/blob/master/docs/useBattery.md) — tracks device battery state. 
- [ ] [useGeolocation](https://github.com/streamich/react-use/blob/master/docs/useGeolocation.md) — tracks geo location state of user's device. 
- [ ] [useHover](https://github.com/streamich/react-use/blob/master/docs/useHover.md) and useHoverDirty — tracks mouse hover state of some element. 
- [ ] [useHash](https://github.com/streamich/react-use/blob/master/docs/useHash.md) — tracks location hash value. 
- [ ] [useIdle](https://github.com/streamich/react-use/blob/master/docs/useIdle.md) — tracks whether user is being inactive.
- [ ] [useIntersection](https://github.com/streamich/react-use/blob/master/docs/useIntersection.md) — tracks an HTML element's intersection. 
- [ ] [useKey](https://github.com/streamich/react-use/blob/master/docs/useKey.md), useKeyPress, useKeyboardJs, and useKeyPressEvent — track keys. 
- [ ] [useLocation](https://github.com/streamich/react-use/blob/master/docs/useLocation.md) and useSearchParam — tracks page navigation bar location state.
- [ ] [useLongPress](https://github.com/streamich/react-use/blob/master/docs/useLongPress.md) — tracks long press gesture of some element.
- [ ] [useMedia](https://github.com/streamich/react-use/blob/master/docs/useMedia.md) — tracks state of a CSS media query. 
- [ ] [useMediaDevices](https://github.com/streamich/react-use/blob/master/docs/useMediaDevices.md) — tracks state of connected hardware devices.
- [ ] [useMotion](https://github.com/streamich/react-use/blob/master/docs/useMotion.md) — tracks state of device's motion sensor.
[...]

PR #67, opened 2021-05-16
feat: new hook `useResizeObserver`

xobotyi:
## What new hook does?

Invokes a callback whenever ResizeObserver detects a change to target's size.

## Checklist

  - #33 

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,14 @@ import { useMountEffect } from "@react-hookz/web/esnext";
   - [**`useSyncedRef`**](https://react-hookz.github.io/?path=/docs/miscellaneous-usesyncedref)
     — Like `useRef`, but it returns immutable ref that contains actual value.
 
-- #### Side-effects
+- #### Side-effect
 
-  - [**`useLocalStorageValue`**](https://react-hookz.github.io/?path=/docs/side-effects-uselocalstoragevalue)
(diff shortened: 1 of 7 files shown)


JoeDuncko (commented):
Questions:

- Why use this over something like https://www.npmjs.com/package/react-resize-detector ?
- Why is this under "Sensor"?
- Why implement this as an observer instead of as a state?

JoeDuncko (approved):
Approved after looking at useMeasure. Note my questions above.

xobotyi:
> * Why use this over something like https://www.npmjs.com/package/react-resize-detector ?
1. First time hearing about it
2. Loking at its code - it is super overcomplicated
3. It uses new observers for each hook, that solution is way less performant than ours: https://groups.google.com/a/chromium.org/g/blink-dev/c/z6ienONUb5A/m/F5-VcUZtBAAJ
4. It has built-in throttlers\debouncers - it is not so versatile

> * Why is this under "Sensor"?
Where it should be? It acts like a sensor - something changes (in terms of element size or position) - it triggers.

> * Why implement this as an observer instead of as a state?
Thus it is more versatile, as for me - im using this hook instead of `useMeasure` for one of my projects because i dont need to rerender component on measurments reading.

PR #68, opened 2021-05-16
feat: new hook `useTitle`

xobotyi:
## What new hook does?

Sets title of the page.

## Checklist

  - #33 

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -34,3 +34,6 @@ export { useSessionStorageValue } from './useSessionStorageValue';
 
 // Sensor
 export { useResizeObserver, IUseResizeObserverCallback } from './useResizeObserver';
+
+// Dom
+export { useTitle, IUseTitleOptions } from './useTitle';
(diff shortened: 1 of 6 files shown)


JoeDuncko (commented):
The example doesn't seem to be working for me
Screen Recording 2021-05-25 at 4 15 07 PM

JoeDuncko:
Also, I think we should include a link to https://github.com/nfl/react-helmet in case they want to use something more powerful. Maybe we should come up with a formal way to include alternatives in our docs?

xobotyi:
> The example doesn't seem to be working for me
> Screen Recording 2021-05-25 at 4 15 07 PM

It actually works, buuuut, the document page is inside iframe =\
Im digging into changing the docs generator to smth else.

JoeDuncko:
That probably makes that a bad example lol. Anything we can do to make it better?

xobotyi:
Nothing while using storybook, attempt to make it handle iframes will be quite bad idea, since react render flows are not aligned between parrent and child frame.

xobotyi:
@JoeDuncko added notification about not working story.

septs:
i want's the hook can rename to `useDocumentTitle`, thanks

xobotyi:
Actually this is a good idea

PR #75, opened 2021-05-19
Remove `types` field in package.json

axelboc:
The `types` folder is not published. I think this should do it.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,4 +2,3 @@
 !/cjs
 !/esm
 !/esnext
-!/types
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "esnext"
   ],
   "main": "cjs/index.js",
-  "types": "types/index.d.ts",
   "esnext": "esm/index.js",
   "scripts": {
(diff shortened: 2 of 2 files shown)


axelboc:
Not sure if `files` in `package.json` is needed at all, since there's already an `.npmignore` file. Feel free to disregard this PR if that's the case.

xobotyi:
Woops =) 
But the issue is different. `types` folder not generated anymore, per-file `.d.ts` files used instead. `types` field of `package.json` and `!/types` line in `.npmignore` is obsolete and should be removed. If not difficult - do it please.

`files` directive used in some IDEs and bundlers so it is better to have it.

xobotyi:
@axelboc it does not, in case of absense of `types` field it searches typings right next to imported files.

Separate `types` folder breaks the hint about multiversion (cjs,esm,esnext) for some IDEs (idea 4ex.)

As for now you have an error since there is no `types` folder.

axelboc:
Right, I get you, thanks for the explanation 😊

Issue #86, opened 2021-05-26
 Wrong import path


nuanyang233:
### What is the current behavior?

build error

```
ModuleNotFoundError: Module not found: Error: Can't resolve './useTitle.stories' in '/web/stories/Dom'
```

### Steps to Reproduce

run `storybook/watch`

### What is the expected behavior?

build work fine

### Environment Details

- _`@react-hookz/web` version:_ 
- _`react` version:_ 
- _`react-dom` version:_ 
- _`typescript` version:_ 
- _OS:_ 
- _Browser:_ 
- _Did this work in previous versions?_

PR #87, opened 2021-05-26
fix: wrong import path

nuanyang233:
- [🐛 Bug Fix](https://github.com/react-hookz/web/issues/86)


xobotyi:
I've already fixed it =)
But nevertheless thanks for participation

xobotyi:
close: #87

Issue #89, opened 2021-05-26
Combine sources, tests, and docs in one place


xobotyi:
As for now, sources, docs and tests are living in separate directories and it is not so great experience to jub all around the project.

Idea is change structure from

```
- src
- - index.ts
- - useState.ts

- tests
- - dom
- - - useState.test.ts
- - ssr
- - - useState.test.ts

- stories
- - State
- - - useState.stories.tsx
- - - useState.story.mdx
```
to
```
- src
- - index.ts
- - useState
- - - useState.ts
- - - useState.test.dom.ts
- - - useState.test.ssr.ts
- - - useState.stories.tsx
- - - useState.story.mdx
```

PR #94, opened 2021-05-28
Combine sources and tests directories

xobotyi:
Implemetation of #89 

Resulting structure is kind of differs, but it is still better than current top-lefel directories separation.


xobotyi:
Documentration merge is postponed till docs engine switch

xobotyi:
As a result of my today digging - there is no storybook alternative that meets our requirements =(
therefore also moved the docs

PR #101, opened 2021-06-01
feat: new hook `useValidator`

xobotyi:
## What new hook does?

Performs validation when any of provided dependencies has changed.

## Checklist

  - #33


src/useValidator/__docs__/example.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+    [text]
+  );
+
+  // validity state type if inferred from validator
+  const [validity] = useValidator(validator, [validator]);
+
+  return (
+    <div>
+      <div>
+        In case below input has an odd number of characters - it is valid, otherwise, the error
```
JoeDuncko, line 33: This isn't exactly clear. Let's reword it to something more like "The input below is only valid if it has an odd number of characters".

src/useValidator/__docs__/example.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+        isValid,
+        error: !isValid ? new Error('text length should be an odd length') : undefined,
+      });
+    },
+    150,
+    [text]
+  );
+
+  // validity state type if inferred from validator
+  const [validity] = useValidator(validator, [validator]);
```
JoeDuncko, line 28: I don't know how I feel about this one. It almost feels like the better UX here would to be to either make this into useValidatorState, or to at least pass the value you are validating to useValidator.

Also, how it's set up here makes it so useValidator seems to be be pretty useless if you weren't using `useDebounceCallback`. You could otherwise just validate on input or on render, which both seem like better practices. Maybe I'm missing something though.

JoeDuncko (approved):
Seems to work fine, but I'm not sold on this implementation.

src/useValidator/__docs__/example.stories.tsx
xobotyi: > to either make this into useValidatorState
It will be implemented as a separate hook later.

> at least pass the value you are validating to useValidator
This will strict you to use validated value only, but what if you want to validate multiple values simultaneously? How to pass them to the validator? Dependencies list is an array, attempt to determine values depending on positions can cause inambiguos bugs.

> seems to be be pretty useless if you weren't using useDebounceCallback.
Althoug it supports synchronous validation - it was designed mostly for async\deferred validation.

> You could otherwise just validate on input or on render, which both seem like better practices.
On-input validation makes multivalue validation very complicated complicated.  
On-render validation forces you to create validity storage by hand and does validation on ach render instead of on-change.

JoeDuncko: `it was designed mostly for async\deferred validation.`

How about we rename this hook to `useAsyncValidator` then? Or `useValidatorEffect`? I think that'd convey the intended use case better.

xobotyi: It is state, not the effect, also it is not the async - only, as for my code I have several synchronous validators.

JoeDuncko: Gotcha. I'm okay rolling with this for now, but I'm still not sold on it.

JoeDuncko (approved):

PR #116, opened 2021-06-10
feat: new hook `useMediaQuery`

xobotyi:
## What new hook does?

Tracks the state of CSS media query.

## Checklist

  - #33 


JoeDuncko (commented):
Question: I don't see any tests that cover when the user puts in invalid media queries. What happens when that's the case? I wonder if there is a better API we can have here to guide the user to write a correctly formatted media query and give better errors when a media query is malformed. What do you think?

JoeDuncko:
However, it does seem to work, so I'm going to approve it.

JoeDuncko (approved):

xobotyi:
@JoeDuncko ideally we can compare passed query and the one that returned in `MediaQueryList`, but it needs way deeper invesigation across browsers.

Ill merge it this way if question will rise anytime later we will do improvement pr.

PR #118, opened 2021-06-12
feat: new hook `useThrottleCallback`

xobotyi:
## What new hook does?

Makes passed function throttled, otherwise acts like `useCallback`.

## Checklist

  - #33 


src/useThrottleCallback/__docs__/example.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+    (ev) => {
+      setState(ev.target.value);
+    },
+    500,
+    []
+  );
+
+  return (
+    <div>
+      <div>Below state will updated no more than once every 500ms</div>
```
JoeDuncko, line 17: will update*

src/useThrottleCallback/__docs__/story.mdx
```diff
@@ -0,0 +1,35 @@
...
+import { Example } from './example.stories';
+
+<Meta title="Callback/useThrottleCallback" component={Example} />
+
+# useThrottleCallback
+
+Makes passed function throttled, otherwise acts like `useCallback`.  
+[\[What is throttling?\]](https://css-tricks.com/debouncing-throttling-explained-examples/#throttle)
+
+> The third argument is dependencies list on `useEffect` manner, passed function will be re-wrapped
```
JoeDuncko, line 11: in `useEffect` manner*

src/useThrottleCallback/__docs__/story.mdx
```diff
@@ -0,0 +1,35 @@
...
+> when delay or dependencies has changed. Changed throttle callbacks still has same timeout, meaning
```
JoeDuncko, line 12: still have*

JoeDuncko (changes requested):
The demo seems to be broken, so I think the throttle implementation might be wrong or I might be misunderstanding using throttle in this case.

It seems that this throttle callback drops the final run on the floor. See the example below - there is more text in the box than the displayed value, even through 500ms has passed.

image

xobotyi:
Ah🤔 interesting, I'll watch other implementations, but this seems to be proper behavior, according to what is throttling.

JoeDuncko:
Just checked, `react-use`'s implementation doesn't drop the last value on the floor, see https://streamich.github.io/react-use/?path=/story/side-effects-usethrottle--demo .

JoeDuncko (approved):
Much better!

PR #130, opened 2021-06-16
fix: rename useThrottleCallback and useDebounceCallback

xobotyi:
### What is the current behavior, and the steps to reproduce the issue?
`useDebounceCallback` and `useThrottleCallback` are not correctly named.

### What is the expected behavior?
It should be `useDebouncedCallback` and `useThrottledCallback`

### How does this PR fix the problem?
Renames tese hooks and ther import\use occurrences.

## Checklist

- [x] Have you read [contribution guideline](../../CONTRIBUTING.md)?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Is there an existing issue for this PR?
  - #129
- [x] Have the files been linted and formatted?
- [x] Have the docs been updated to match the changes in the PR?
- [x] Have the tests been updated to match the changes in the PR?
- [x] Have you run the tests locally to confirm they pass?

PR #131, opened 2021-06-16
feat: add `maxWait` parameter to `useDebouncedCallback` hook

xobotyi:
`maxWait` defines maximum amount of milliseconds that function can be delayed before it's force execution. 0 means no max wait (default).

**should be merged after #130**

## Checklist

  - _link issue here_

### DIFF
--- a/src/useDebouncedCallback/useDebouncedCallback.ts
+++ b/src/useDebouncedCallback/useDebouncedCallback.ts
@@ -1,40 +1,85 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { DependencyList, useMemo, useRef } from 'react';
+import { useUnmountEffect } from '../useUnmountEffect/useUnmountEffect';
+
+export interface IDebouncedFunction<Args extends any[], This> {
+  (this: This, ...args: Args): void;
+}
 
 /**
  * Makes passed function debounced, otherwise acts like `useCallback`.
  *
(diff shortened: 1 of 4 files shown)


JoeDuncko:
While we are here, should we also implement `leading` and `trailing` options, to copy [lodash's implementation](https://lodash.com/docs/4.17.15#debounce) ?

src/useDebouncedCallback/useDebouncedCallback.ts
```diff
@@ -0,0 +1,85 @@
...
+  (this: This, ...args: Args): void;
+}
+
+/**
+ * Makes passed function debounced, otherwise acts like `useCallback`.
+ *
+ * @param callback Function that will be debounced.
+ * @param delay Debounce delay.
+ * @param deps Dependencies list when to update callback.
+ * @param maxWait Maximum amount of milliseconds that function can be delayed
```
JoeDuncko, line 15: I like Lodash's description of `maxWait`

`The maximum time func is allowed to be delayed before it's invoked.`

xobotyi:
Leading and trailing is only a matter of single implementation for both, throttle and debounce (if you look their throttle - it is only rewrapped debounce).
I prefer more specialized and therefore short and simple implementations.

JoeDuncko (approved):
LGTM, but see comments above

PR #132, opened 2021-06-16
feat: change args for `useThrottledCallback` and `useDebouncedCallback`

xobotyi:
`useThrottledCallback` and `useDebouncedCallback` has different arguments order comparing to `useCallback` (delay argument is before deps list). It is mentally harder to handle such differences.

BREAKING CHANGE:
`delay` and `deps` arguments are swapped position for `useThrottledCallback` and `useDebouncedCallback` hooks to be aligned with `useCallback` signature.

**merge after #131**

## Checklist

  - _link issue here_

### DIFF
--- a/src/useDebouncedCallback/__docs__/example.stories.tsx
+++ b/src/useDebouncedCallback/__docs__/example.stories.tsx
@@ -8,14 +8,14 @@ export const Example: React.FC = () => {
     (ev) => {
       setState(ev.target.value);
     },
-    300,
     [],
+    300,
     500
   );
 
   return (
     <div>
-      <div>Below state will update 200ms after last change, but at least once every 500ms</div>
+      <div>Below state will update 300ms after last change, but at least once every 500ms</div>
       <br />
       <div>The input`s value is: {state}</div>
       <input type="text" onChange={handleChange} />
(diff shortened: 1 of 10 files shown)


JoeDuncko (approved):

JoeDuncko (changes requested):
Actually, I think I'm less convinced that the dependency list shouldn't be the last argument. How do other libraries handle this?

Thinking about this more, I'm wondering if we should move to an `options` object here like [lodash](https://lodash.com/docs/4.17.15#debounce).

xobotyi:
Disagree about options object, since at leas it is longer record:
```ts
useDebouncedEffect(()=>{}, [], 200, 600);
```
vs
```ts
useDebouncedEffect(()=>{}, [], { delay: 200, msxWait: 600 });
```
Options object is more obvious though, but most IDE highlight arguments so this should not be an issue.

If there would be more options - i would agree, but at leas for now, on my taste, arguments are better solution

xobotyi:
> How do other libraries handle this?

Have no idea, havent watched any other solutions =)

JoeDuncko (approved):
kk, you've convinced me. Until we have more args this should be fine.

PR #155, opened 2021-06-24
feat: new hook `useMap`

xobotyi:
## What new hook does?

Tracks the state of a `Map`.

## Checklist

  - #33

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,8 @@ import { useMountEffect } from "@react-hookz/web/esnext";
 
   - [**`useDebouncedState`**](https://react-hookz.github.io/web/?path=/docs/state-usedebouncedstate)
     — Lise `useSafeState` but its state setter is debounced.
+  - [**`useMap`**](https://react-hookz.github.io/web/?path=/docs/state-usemap)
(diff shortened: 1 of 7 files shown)


JoeDuncko (approved):
LGTM

DanHarman:
I don't understand why this has a mutable object api in contrast to useList which returns the operations distinct to the list. This makes the library inconsistent?

xobotyi:
1) Сopying a map of 1k numbers is two orders of magnitude slower than copying same size array (like it is slow af).
2) There is no way of giving an access to a map without making it mutable, `Map` method mutates its source and there is nothing to do with it.

DanHarman:
Understood. I wonder if it would be better to return the proxied apis in the way list does, although then risk of people mutating the map directly without the proxied versions unless they are removed from the returned map. All quite difficult trade offs so I understand your decision now.

DanHarman:
Understood. I wonder if it would be better to return the proxied apis in the way list does, although then risk of people mutating the map directly without the proxied versions unless they are removed from the returned map. All quite difficult trade offs so I understand your decision now.

PR #386, opened 2021-10-09
feat(useFavicon): port `useFavicon` from `react-use`

alexnaiman:
## Checklist

  - #33 


alexnaiman:
@septs If the pull request is merged/accepted, could you please add the `hacktoberfest-accepted` label?

ArttuOll:
I noticed that this PR is quite old, so I figured it was time to give it some love.

I went over this PR and fixed things that needed to be fixed in order for this to be merged (added missing documentation, improved example, refactored code a little). Let's review this and get this one merged!

xobotyi:
Ah, about this one. i'm still making my mind, but tupping to decision not to implement hooks that modify <head> contents since it is more complicated that just throwing elements there. There is a better library called react-helmet and react-helmet-async doing this job pretty well, and, in most cases, even better and in more versatile than any hook possible.

ArttuOll:
> Ah, about this one. i'm still making my mind, but tipping to decision not to implement hooks that modify contents since it is more complicated that just throwing elements there. There is a better library called react-helmet and react-helmet-async doing this job pretty well, and, in most cases, even better and in more versatile than any hook possible.

Just had a look at react-helmet-async, and fair enough, there seems to be more to the SSR side than I imagined.

Could you write your thoughts like this to the pull requests in the future? That way we would all be on the same page and unnecessary work could be avoided.

Should we close this one then and add a mention to the docs that this one will not be implemented?

ArttuOll:
> but tipping to decision not to implement hooks that modify <head> contents

What are going to do with `useDocumentTitle` then? Deprecate it or keep it?

xobotyi:
1) this hook has to be rejected
2) we have to remove hooks modifying `<head>` contents (yes, it is a breaking change)
3) we have to add clarification on migration guide (docs) with suggestions to use `react-helmet` or `react-helmet-async`
4) we have to add clarification notes to issue with migration list

ArttuOll:
> 1. this hook has to be rejected
> 2. we have to remove hooks modifying `<head>` contents (yes, it is a breaking change)
> 3. we have to add clarification on migration guide (docs) with suggestions to use `react-helmet` or `react-helmet-async`
> 4. we have to add clarification notes to issue with migration list

Sounds good!

PR #530, opened 2021-12-14
feat: new hook useGetSet

fengkx:
## What new hook does?
Port of `useGetSet` in react-use
## Checklist

   #33

### DIFF
--- a/src/useGetSet/__tests__/dom.ts
+++ b/src/useGetSet/__tests__/dom.ts
@@ -1,0 +1,62 @@
+import { renderHook, act } from '@testing-library/react-hooks/dom';
+import { useGetSet } from '../..';
+
+const setUp = (initialValue: any) => renderHook(() => useGetSet(initialValue));
+
+beforeEach(() => {
+  jest.useFakeTimers();
+});
+
+describe('useGetSet', () => {
+  it('should be defined', () => {
+    expect(useGetSet).toBeDefined();
+  });
+
+  it('should render', () => {
+    const { result } = setUp('foo');
+    expect(result.error).toBeUndefined();
+  });
+
+  it('should init getter and setter', () => {
+    const { result } = setUp('foo');
+    const [get, set] = result.current;
+    expect(get).toBeInstanceOf(Function);
+    expect(set).toBeInstanceOf(Function);
+  });
+
+  it('should get current value', () => {
+    const { result } = setUp('foo');
+    const [get] = result.current;
+    expect(get()).toBe('foo');
+  });
+
+  it('should set new value', () => {
+    const { result } = setUp('foo');
+    const [get, set] = result.current;
+    act(() => set('bar'));
+    const currentValue = get();
+    expect(currentValue).toBe('bar');
+  });
+
+  it('should get and set expected values when used in nested functions', () => {
+    const { result } = setUp(0);
+    const [get, set] = result.current;
+    const onClick = jest.fn(() => {
+      setTimeout(() => {
+        set(get() + 1);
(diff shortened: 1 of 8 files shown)


src/useGetSet/useGetSet.ts
```diff
@@ -0,0 +1,19 @@
...
+  const ref = useRef<S>(resolveHookState(initialState));
+  const get = useCallback(() => ref.current, []);
+  const set = useCallback(
+    (nextState: INextState<S>) => {
+      ref.current = resolveHookState(nextState, ref.current);
+      rerender();
+    },
+    [rerender]
+  );
+  return [get, set];
```
xobotyi, line 18: Why not to use `useSafeState`, `useSyncedRef` and `useCallback`?

Besides this code will work - id prefer `native` state holder

xobotyi (changes requested):

src/useGetSet/useGetSet.ts
fengkx: Sorry, I don't understand here.


> id prefer `native` state holder

what is it means here

fengkx: either useState nor useSafeState update the value immediately. That why this hook is need and `useRerender` is used in setter to control UI updates.

How `useSyncedRef` can be used here

xobotyi: This implementation breaks `state flow` of react, and will break in strict mode, 4ex.

What i propose is to implement hook like this:

```ts
export function useGetSet<S>(
  initialState: IInitialState<S>
): [get: () => S, set: Dispatch<SetStateAction<S>>] {
  const [state, setState] = useSafeState(initialState);
  const stateRef = useSyncedRef(state);

  return [useCallback(() => stateRef.current, []), setState];
}
```

fengkx: This implementation doesn't return latest value in getter. /

xobotyi: Well it shouldn't  - hook description says that it is like `useState`, but there is not a word about differenet behavior.

Therefore i assume that it should work exactly like `useState`, but return state getter instead of state itself.
I dont see any reason to deviate from react's state commitment flow. Or i simply dont know usecase for your implementation.

fengkx: make sense to me

fengkx: But how to write test, since I don't know when will react state change

xobotyi: Just wrap all `set` calls with `act` method of testing library.

xobotyi: Also, PR lack of readme changes.

fengkx: > Just wrap all `set` calls with `act` method of testing library.

```
  it('should set new value', () => {
    const { result } = setUp('foo');
    const [get, set] = result.current;
    act(() => set('bar'));
    const currentValue = get();
    expect(currentValue).toBe('bar');
  });
```
This test cannot pass /

src/useGetSet/__tests__/dom.ts
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    // simulate 3 clicks
+    onClick();
+    onClick();
+    onClick();
+
+    act(() => {
+      jest.runAllTimers();
+    });
```
xobotyi, line 57: it is way overcomplicated, why not to simply set state inside act?
```
act(()=>{
 set();
 set();
 set();
})

PR #562, opened 2022-01-08
docs(useDefault): Suggest `useMediatedState` instead 

AndreasNel:
## What new hook does?

It works like `useSafeState`, except it will return the provided default value when the state is set to `undefined` or `null`.

## Checklist

  - #33 

### DIFF
--- a/src/__docs__/migrating-from-react-use.story.mdx
+++ b/src/__docs__/migrating-from-react-use.story.mdx
@@ -567,7 +567,18 @@ No plans to implement
 
 #### useDefault
 
-Not implemented yet
+No plans to implement, rather use [useMediatedState](/docs/state-useMediatedState--example) like
+so:
+
+```javascript
+const initialValue = 'world';
+const defaultValue = 'you';
+const [greeting, setGreeting] = useMediatedState(
+  initialValue,
+  (newValue) => newValue ?? defaultValue
+);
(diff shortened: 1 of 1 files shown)


xobotyi:
TBH i dont think we need that hook since it is easily (and versatile) achievable via `useMediatedState` hook.

@JoeDuncko WDYT?

JoeDuncko:
I wonder if we should have a cookbook of common hooks that can be easily derived from our hooks, or something like that.

Maybe as examples in Storybook?

I know that doesn't exactly answer the question though.

xobotyi:
I've been thinking about that too (cookbook). Maybe separte page near migration and reference in mugration like `we dont have this hook implemented directly as it can be easily achieved via other hook, and here is example link.`

JoeDuncko:
> Yeah a cookbook sounds good, do you want me to create such a page? Or would you prefer doing it?

Sorry for the slow reply. If you know your way around Storybook, go for it! I'm still new to Storybook, so I'd have to figure out the best way to present alternative examples. Ideas welcome.

AndreasNel:
Apologies for leaving this for so long, life kind of got in the way and this got away from me. Please check if the documentation suggestion is correct.

xobotyi:
LGTM, merging after all check passed.

PR #583, opened 2022-01-14
Create .prettierrc.js

kylemh:
See: https://github.com/react-hookz/eslint-config/pull/2 for context!

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,12 @@
+module.exports = {
+  printWidth: 100,
+  tabs: false,
+  semi: true,
+  singleQuote: true,
+  trailingComma: 'es5',
+  endOfLine: 'lf',
+  tabWidth: 2,
+  bracketSpacing: true,
+  bracketSameLine: true,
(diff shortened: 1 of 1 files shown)


.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
xobotyi, line 1: What if:
```
const eslintConfig = require("@react-hookz/eslint-config/.prettierrc.js")

module.exports = {
  ...eslintConfig,
};
```

This way it inherits the config used in eslint and easily can be extended

kylemh: Good idea! I can also put it into `package.json` like you said on Discord, but the config file seems to follow what's been done with other similar tools.

Let me know!

xobotyi: i'd stay with separate file.

kylemh:
Done. Obviously, we'll want to update the version of the eslint-config package and upgrade here before merging.

kylemh:
Whoops. Deleted fork.

PR #623, opened 2022-01-31
use-raf hook: animation hook for react using requestAnimationFrame

Sparkenstein:
## Checklist

  - NOTE: need more tests

### DIFF
--- a/src/useRaf/useRaf.ts
+++ b/src/useRaf/useRaf.ts
@@ -1,0 +1,46 @@
+import { useState } from 'react';
+import { useIsomorphicLayoutEffect } from '..';
+
+/**
+ *
+ * @param ms milliseconds for how long to keep rendering component
+ * @param delay delay in milliseconds after which to start re-rendering component
+ * @returns number
+ */
+export function useRaf(ms = 1e12, delay = 0): number {
+  const [elapsed, setElapsed] = useState<number>(0);
+
+  useIsomorphicLayoutEffect(() => {
+    let frame: number;
+    let stop: NodeJS.Timeout;
+    let start: number;
+
+    const onFrame = () => {
+      const time = Math.min(1, (Date.now() - start) / ms);
+      setElapsed(time);
+      loop();
+    };
+
+    function loop() {
+      frame = requestAnimationFrame(onFrame);
+    }
+
+    const onStart = () => {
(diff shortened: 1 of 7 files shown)


Sparkenstein:
Few questions:

 - I am not really good with Jest, need to add more tests. please suggest new tests and approaches.
 - I was referring react-use test cases for useRaf, it uses [raf-stuf](https://github.com/streamich/react-use/blob/master/tests/useRaf.test.ts) which is written in JS. so couldn't use it in tests as types are not available for it

Sparkenstein:
any plan to merge this?

xobotyi:
i have several notices on implementation, but I'm ill ATM, will do the review later.

src/useRaf/useRaf.ts
```diff
@@ -0,0 +1,46 @@
+import { useState } from 'react';
+import { useIsomorphicLayoutEffect } from '..';
+
+/**
+ *
+ * @param ms milliseconds for how long to keep rendering component
+ * @param delay delay in milliseconds after which to start re-rendering component
+ * @returns number
+ */
+export function useRaf(ms = 1e12, delay = 0): number {
```
xobotyi, line 10: `ms` should have no default value, also, it is better to rename in to `duration`

xobotyi, line 10: 1. This hook should have handles to stop and start re-rendenring process, start should apply optional rerender duration.
2. In implementation it should reuse `useRafCallback` hook, that is already implemented in this package.
3. `elapsed` should use `useSafeState`, since it is updated asynchronously - there is a chance, that component will be unmounted before state update.

xobotyi (changes requested):

xobotyi:
> I was referring react-use test cases for useRaf, it uses [raf-stub](https://github.com/streamich/react-use/blob/master/tests/useRaf.test.ts) which is written in JS. so couldn't use it in tests as types are not available for it
Tests don't require `raf-stub` since Jest provides pretty much all needed functionality, see https://github.com/react-hookz/web/blob/master/src/useRafCallback/__tests__/dom.ts

SCG82:
Please consider https://github.com/streamich/react-use/pull/2215

xobotyi:
Closing due t lack of response from author

PR #654, opened 2022-02-16
docs: adds documentation on how to migrate from useSetState hook

punkle:
adds documentation on hot to migrate from useSetState hook

## Checklist

  - #33
- [ ] Have the tests been added to cover new hook?

### DIFF
--- a/src/__docs__/migrating-from-react-use.story.mdx
+++ b/src/__docs__/migrating-from-react-use.story.mdx
@@ -603,7 +603,25 @@ Not implemented yet
 
 #### useSetState
 
-Not implemented yet
+`React.useState` maybe used to achieve the same with minimal changes.
+
+OLD in `react-use`:
+
+```javascript
+const [state, setState] = useSetState({});
+
+setState({ hello: 'world' });
+setState({ foo: 'bar' });
+```
+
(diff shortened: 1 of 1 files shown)


xobotyi:
@JoeDuncko should we even implement it? 

It is slightly shorter version of `setState((state)=>{...state, newItem})` and i dont see any need to make unnecessary(imho) wrapper.

JoeDuncko:
@xobotyi Yeah, I agree. Considering this is a one liner otherwise, I don't see the value to adding it to `hookz`. I can be convinced though. Thoughts @punkle ?

I wonder if we should go through the `Not Yet Implemented` list and remove some. Normally 

That being said, thank you for your contribution @punkle . Even if we don't end up using it, I appreciate you making such a thorough PR. Like I said, I can definitely be convinced that this has value over `setState((state)=>{...state, newItem})`, I'm just not seeing it on first glance.

punkle:
Hey. I am happy to leave it once it's possible somehow. I had been attempting to port a downstream project that was using this function. If you agree I'll reduce this change to documentation about how to migrate from using react-use useSetState.

punkle:
Thanks for the comments, that's done.

PR #751, opened 2022-05-11
feat(useCounter): port `useCounter` from `react-use`

ArttuOll:
## What new hook does?
This hook tracks a numeric value and offers functions for manipulating it.

The exact same API from `react-use` is retained.

## Checklist

- [X] Does the code have comments in hard-to-understand areas?
  - #33 
- [X] Have the tests been added to cover new hook?

### DIFF
--- a/src/useCounter/useCounter.ts
+++ b/src/useCounter/useCounter.ts
@@ -1,6 +1,6 @@
-import { SetStateAction, useCallback, useRef } from 'react';
+import { SetStateAction, useMemo } from 'react';
+import { useMediatedState, useSyncedRef } from '..';
 import { IInitialState, resolveHookState } from '../util/resolveHookState';
-import { useSafeState, useSyncedRef } from '..';
 
 export interface CounterActions {
   /**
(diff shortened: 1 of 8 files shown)


src/useCounter/useCounter.ts
```diff
@@ -0,0 +1,116 @@
...
+      }
+
+      if (typeof min !== 'undefined') {
+        fittedValue = Math.max(min, fittedValue);
+      }
+
+      return fittedValue;
+    },
+    [max, min]
+  );
```
xobotyi, line 70: It could've been done via `useMediatedState`. Also there would be no need in `useCallback`, as `useMediatedState` perfectly handles dynamic mediators.

xobotyi (changes requested):

src/useCounter/useCounter.ts
ArttuOll: Interesting! Should get more experience using this library in order to be able to think more through all these hooks.

I see you already committed a change where you implemented this. Is there anything I should still do?

xobotyi: No, i'll merge today.

PR #883, opened 2022-07-22
feat: Implement useCopyToClipboard hook

pbellon:
## What new hook does?
New hook inspired by react-use's `useCopyToClipboard`, copies some text to clipboard. It returns an array composed as follow:
- a `copyState` that reflects copy process (if successful, erroneous and what data has been copied),
- `copyToClipboard`, main callback to copy a string into clipboard
- `resetCopyState`, new callback not present in react-use's implementation to easily reset `copyState` to initial value (all fields undefined)

Countrary to react-use's implementation, this hook does not rely on `copy-to-clipboard` external dependency. Instead it simply uses [navigator.clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) which seems to have good support on major browser vendors (see [caniuse compatibly table](https://caniuse.com/?search=navigator.clipboard)). The `noUserInteraction: boolean` has been removed from `copyState` because it seemed specific to `copy-to-clipboard` usage.

Also added a `success: boolean` field in `copyState` to ease usage but this is up to debate since we could do as react-use's implementation and only rely on `value` not being undefined.

## Checklist

  - #33 (useCopyToClipboard)

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Coming from `react-use`? Check out our
     — Provides vibration feedback using the Vibration API.
   - [**`usePermission`**](https://react-hookz.github.io/web/?path=/docs/navigator-usepermission--example)
     — Tracks a permission state.
+  - [**`useCopyToClipboard`**](https://react-hookz.github.io/web/?path=/docs/navigator-usecopytoclipboard--example)
 
 - #### Miscellaneous
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -50,6 +50,7 @@ export {
 } from './useValidator/useValidator';
 
 // Navigator
+export { useCopyToClipboard, CopyToClipboardState } from './useCopyToClipboard/useCopyToClipboard';
 export {
   useNetworkState,
   IUseNetworkState,
--- a/src/useCopyToClipboard/__docs__/example.stories.tsx
+++ b/src/useCopyToClipboard/__docs__/example.stories.tsx
@@ -1,0 +1,48 @@
+import * as React from 'react';
+import { useState } from 'react';
+import { useCopyToClipboard } from '../useCopyToClipboard';
+
+export const Example: React.FC = () => {
+  const [value, setValue] = useState('');
+
+  const [copyState, copyToClipboard, resetCopyState] = useCopyToClipboard();
+
+  return (
+    <div>
+      <div style={{ marginBottom: '1em' }}>
+        <label htmlFor="value-to-copy">
+          <div>Enter value to copy</div>
+          <input id="value-to-copy" value={value} onChange={(e) => setValue(e.target.value)} />
+        </label>
+      </div>
+      <div style={{ marginBottom: '0.5em' }}>
+        <button disabled={value.length === 0} onClick={() => copyToClipboard(value)}>
+          Copy to clipboard
+        </button>
+        <button onClick={() => copyToClipboard(1234 as unknown as string)}>
+          Copy invalid data to clipboard (error)
+        </button>
+        <button onClick={resetCopyState}>Reset copy state</button>
+      </div>
+      <div style={{ marginBottom: '1em' }}>
+        <pre>
+          <code>
+            {JSON.stringify({
+              ...copyState,
+              error: copyState.error ? { message: copyState.error.message } : undefined,
+            })}
+          </code>
+        </pre>
+      </div>
+      {copyState.success ? (
+        <div style={{ color: 'green' }}>Successfully copied to clipboard</div>
+      ) : copyState.success === false ? (
+        <div style={{ color: 'red' }}>
+          Failed to copy to clipboard: {copyState.error?.message ?? 'No error set'}
+        </div>
+      ) : (
+        <div>Not copied to clipboard yet</div>
+      )}
+    </div>
+  );
+};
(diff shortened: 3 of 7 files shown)


xobotyi (changes requested):
1) due to component lack of clipboard read - i really dont seee the need in state at all.

    IMO there is only need in returned callback that returns copied text or throw an error in case of any. Current Implementation will lead to unnecessary re-renders, as of my experience - i never dont care what been copied - most of the time it is simple success or error message depending on results.
    
    So i'd suggest simplifying the hook, but threfore component would be redundant, as it han be replaced with single function.
  
    Instead of that i can suggest reimplementing this hook as more complex hook `useClipboard` that will give access to clipboard reads and reaction to [clipboard events occured on docment](https://developer.mozilla.org/en-US/docs/Web/API/ClipboardEvent), events list can be configurable for hook to provide maximal granularity of configuration.

2) clipboard reset and clipboard set can be merged into single function - do reset when `undefined` received.

3) copy function can receive any value, as there are loads of stringable objects that are not strings, just perform `String(val)` before passing to clipboard.

4) you did not provide any jsdoc for hook, as it is required by contribution guide.

ArttuOll:
Is this new hook, `useClipboard`, being worked on by anyone at the moment? If not, I could start implementing it over the weekend.

xobotyi:
@ArttuOll you can take and fix my remarks

xobotyi:
@ArttuOll next pick this one when you'll have free time🍻

ArttuOll:
@xobotyi I'm currently sketching the API for `useClipboard`. What do you think about this?

## Return

- **eventHandlers**
    - **onCopy** _`(event: ClipboardEvent) => void`_ - Event handler for the `copy` event.
    - **onCut** _`(event: ClipboardEvent) => void`_ - Event handler for the `cut` event.
    - **onPaste** _`(event: ClipboardEvent) => void`_ - Event handler for the `paste` event.
- **setClipboardContent** _`(content: unknown) => void`_ - Populate the clipboard with `content`.

xobotyi:
im not sure about returned event handlers - i think usage of [ClipboardAPI](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) would be better approach

ArttuOll:
> im not sure about returned event handlers - i think usage of [ClipboardAPI](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API) would be better approach

You mean something like exposing `readText` and `writeText` from `navigato.clipboard` through the hook?

xobotyi:
and ways to assign clipboard events handlers, yes.

ArttuOll:
How about this?

## Return

- **eventHandlers**
    - **onCopy** _`(event: ClipboardEvent) => void`_ - Event handler for the `copy` event.
    - **onCut** _`(event: ClipboardEvent) => void`_ - Event handler for the `cut` event.
    - **onPaste** _`(event: ClipboardEvent) => void`_ - Event handler for the `paste` event.
- **setClipboardContent** _`(content: unknown) => void`_ - Populate the clipboard with `content`.
- **readClipboardContent** _`() => Promise<string>`_ - Read clipboard content
- - -
> and ways to assign clipboard events handlers

I'm confused about this. Does **eventHandlers** not already cover all the possible clipboard events?

Also, should we go with `readText` and support only string data or with `read` with supports non-text content, but isn't supported by Firefox? I would probably go with just `readText`, because `read` seems to be too bleeding edge.

ArttuOll:
Rethinking this now, I think I'll start implementing something like this. We can change it when I submit a PR, if we need to.

## Return
- **write** _`(newClipText: string, onSuccess: () => void, onFailure: () => void ) => void`_ - Write text to the clipboard.
- **read** _`(onSuccess: (clipText: string) => void, onFailure: () => void ) => void`_ - Read text from the clipboard.

xobotyi:
Yes =)
Sorry for long response - been really loadded at work recently.

This is exactly what i've been concerned of - you rprevious aproach offered callbacks that should be passed to desired element what is not necessary needed.

The only concern now - it seems no possbility to somehow validate\filter the target when we're dealing with clipboard, at least im not able to see it in offered signature.

xobotyi:
Given this more thinking lately.
Do we even need this hook? 

Previously such hook been needed due to poor spreading of clipboad api and usage of third-party libs, which is not the case nowadays.
Usually such hook used to copy something to user's clipboard therefore there is no need for state updates since application know the state in advance.

I think i'd vote ditching that hook and add note to migration guide. WEB-API covering whole functionality by itself wouthout need of custom hook.

ArttuOll:
> Given this more thinking lately. Do we even need this hook?
> 
> Previously such hook been needed due to poor spreading of clipboad api and usage of third-party libs, which is not the case nowadays. Usually such hook used to copy something to user's clipboard therefore there is no need for state updates since application know the state in advance.
> 
> @JoeDuncko @ArttuOll I think i'd vote ditching that hook and add note to migration guide. WEB-API covering whole functionality by itself wouthout need of custom hook.

This is pretty much what I thought when I started implementing this. The hook would do pretty much nothing of value, so the hook is unnecessary in my opinion also.

JoeDuncko:
> > Given this more thinking lately. Do we even need this hook?
> > Previously such hook been needed due to poor spreading of clipboad api and usage of third-party libs, which is not the case nowadays. Usually such hook used to copy something to user's clipboard therefore there is no need for state updates since application know the state in advance.
> > @JoeDuncko @ArttuOll I think i'd vote ditching that hook and add note to migration guide. WEB-API covering whole functionality by itself wouthout need of custom hook.
> 
> This is pretty much what I thought when I started implementing this. The hook would do pretty much nothing of value, so the hook is unnecessary in my opinion also.

Sorry for the delay. I agree.

PR #886, opened 2022-07-31
feat(useTimeoutEffect): Implement useTimeoutEffect

AndreasNel:
## What new hook does?
Functions like `setTimeout`, except in hook form. Returns a method to cancel the timeout, and a method to restart the timeout. It does not cause any rerenders.

## Checklist

  - https://github.com/react-hookz/web/issues/33

### DIFF
--- a/src/useTimeoutEffect/__tests__/ssr.ts
+++ b/src/useTimeoutEffect/__tests__/ssr.ts
@@ -1,0 +1,33 @@
+import { renderHook } from '@testing-library/react-hooks/server';
+import { useTimeoutEffect } from '../..';
+
+describe('useTimeoutEffect', () => {
+  beforeAll(() => {
+    jest.useFakeTimers();
+  });
+
+  beforeEach(() => {
+    jest.clearAllTimers();
+  });
+
+  afterAll(() => {
+    jest.useRealTimers();
+  });
+
+  it('should be defined', () => {
+    expect(useTimeoutEffect).toBeDefined();
+  });
+
+  it('should render', () => {
+    const { result } = renderHook(() => useTimeoutEffect(() => {}, 123));
+    expect(result.error).toBeUndefined();
+  });
+
+  it('should not invoke callback after timeout', () => {
+    const spy = jest.fn();
+    renderHook(() => useTimeoutEffect(spy, 100));
+
+    jest.advanceTimersByTime(100);
+    expect(spy).not.toHaveBeenCalled();
+  });
+});
--- a/src/useTimeoutEffect/useTimeoutEffect.ts
+++ b/src/useTimeoutEffect/useTimeoutEffect.ts
@@ -1,0 +1,49 @@
+import { useCallback, useEffect, useRef } from 'react';
+import { useSyncedRef } from '..';
+
+type TimeoutID = ReturnType<typeof setTimeout> | null;
+
+const cancelTimeout = (id: TimeoutID) => {
+  if (id) {
+    clearTimeout(id);
+  }
+};
+
+/**
+ * Like `setTimeout` but in the form of a react hook.
+ *
+ * @param callback Callback to be called after the timeout. Changing this
+ * will not reset the timeout.
+ * @param ms Timeout delay in milliseconds. `undefined` disables the timeout.
+ * Keep in mind, that changing this parameter will re-set timeout, meaning
+ * that it will be set as new after the change.
+ */
+export function useTimeoutEffect(
+  callback: () => void,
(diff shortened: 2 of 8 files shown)


AndreasNel:
I was hesitant about including the `cancel` and `reset` return values. It is possible to get the same functionality by changing the timeout to `undefined` in order to cancel the function, and a `reset` is possible by either changing the timeout, or setting it to `undefined` and then back to the same value if necessary.

The downside of this is that these methods then require rerenders, which the original method didn't need, and the current `cancel` and `reset` methods also don't need in order for them to work correctly.

src/useTimeoutEffect/useTimeoutEffect.ts
```diff
@@ -0,0 +1,46 @@
...
+    return v;
+  });
+
+  const cancel = useCallback(() => {
+    cancelTimeout(timeoutIdRef.current);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const reset = useCallback(() => {
+    if (!ms && ms !== 0) {
```
xobotyi, line 33: Isn't `typeof ms === 'undefined'` better? At least in terms of readability.

src/useTimeoutEffect/useTimeoutEffect.ts
```diff
@@ -0,0 +1,46 @@
...
+      return;
+    }
+    timeoutIdRef.current = setTimeout(() => cbRef.current(), ms);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [ms]);
+
+  useEffect(() => {
+    reset();
+    return cancel;
+  }, [cancel, reset]);
```
xobotyi, line 43: `cancel` dont change over time, why is it here?🤔

src/useTimeoutEffect/useTimeoutEffect.ts
```diff
@@ -0,0 +1,46 @@
...
+const cancelTimeout = (id: TimeoutID) => {
+  if (id) {
+    clearTimeout(id);
+  }
+};
+
+/**
+ * Like `setTimeout` but in the form of a react hook.
+ *
+ * @param callback Callback to be called after the timeout.
```
xobotyi, line 15: It is better to describe that changes in callback do not lead to timeut reset (some developers might expect it).

src/useTimeoutEffect/__tests__/ssr.ts
```diff
@@ -0,0 +1,13 @@
+import { renderHook } from '@testing-library/react-hooks/server';
+import { useTimeoutEffect } from '../..';
+
+describe('useTimeoutEffect', () => {
```
xobotyi, line 4: i'd also add tests to ensure that callback is not invoked within SSR environment.

xobotyi (changes requested):
Overall - good approach on using `useHookableRef` but IMO it is not the best place for it.

Below implementation is sligtly simpler, has less memory footprint and its return is stable across the renders

```ts
export function useTimeoutEffect(callback: () => void, ms?: number): [() => void, () => void] {
  const cbRef = useSyncedRef(callback);
  const msRef = useSyncedRef(ms);
  const timeoutIdRef = useRef<TimeoutID>(null);

  const cancel = useCallback(() => {
    cancelTimeout(timeoutIdRef.current);
  }, []);

  const reset = useCallback(() => {
    if (typeof msRef.current === 'undefined') return;

    cancel();
    timeoutIdRef.current = setTimeout(() => cbRef.current(), msRef.current);
  }, []);

  useEffect(() => {
    reset();
    return cancel;
  }, [ms]);

  return [cancel, reset];
}
````

src/useTimeoutEffect/useTimeoutEffect.ts
AndreasNel: I think I just added it to remove the ESLint error instead of disabling the rule. I'll make the change 👍

src/useTimeoutEffect/__tests__/ssr.ts
AndreasNel: Why should it not be invoked? Does `setTimeout` not work in SSR? I haven't worked with SSR yet, so I don't know about exactly all the details.

xobotyi: Effects dont work in SSR, therefore callback of this hook should not be invoked in SSR. By naming hook as `*Effect` we must comply default behaviour

xobotyi (approved):
Good work, thanks for submitting!

PR #895, opened 2022-08-07
feat: `useCustomCompareMemo` - like useMemo but uses custom deps comparator

MichalTarasiuk:
`useRetain`: useMemo but with a callback to check that factory fn can be invoked


xobotyi (commented):
Shoulnt it be called a `useCustomCompareMemo` and a custom over it `useDeepComapreMemo`? Like `useCustomCompareEffect` and `useDeepCompareEffect`?  
Besides in current implementation comparator is a negtive predicate (which is usually bad pattern).

xobotyi (commented):
Also, why is there changes to yarn.lock?

MichalTarasiuk:
> Also, why is there changes to yarn.lock?

I just run `yarn installl`. Which version node do you have?

xobotyi:
16, but dependencies are maintained via dependabot.
Besides from that I was asking not about origin, but unrelated changes in pr.

You should remove lockfile changes from pr

src/useCustomCompareMemo/useCustomCompareMemo.ts
```diff
@@ -0,0 +1,33 @@
+import { useMemo, useRef } from 'react';
+
+import type { DependencyList } from 'react';
+
+type DepsAreEqual<TDeps extends DependencyList> = (savedDeps: TDeps, deps: TDeps) => boolean;
```
xobotyi, line 5: Why do you use prefixes, btw? What does `T` stands for? It is considered bad practice in TS community nowadays, and described in contribution guide

xobotyi (changes requested):

src/useCustomCompareMemo/useCustomCompareMemo.ts
MichalTarasiuk: In my understanding "T" prefix mean that type is generic, but i respect rules of code. Give me a second, I fix it.

README.md
```diff
@@ -152,6 +152,8 @@ Coming from `react-use`? Check out our
 
   - [**`useSyncedRef`**](https://react-hookz.github.io/web/?path=/docs/miscellaneous-usesyncedref--example)
     — Like `useRef`, but it returns immutable ref that contains actual value.
+  - [**`useCustomCompareMemo`**](https://react-hookz.github.io/web/?path=/docs/miscellaneous-useCustomCompareMemo--example)
+    — Like `useMemo` but with a callback to defines that factory function can be invoked.
```
xobotyi, line 156: Like `useMemo` but uses provided comparator function to validate dependency changes.

src/useCustomCompareMemo/useCustomCompareMemo.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @param factory useMemo factory function
+ * @param deps useMemo dependency list
+ * @param depsAreEqual defines that factory function can be invoked
+ * @returns useMemo result
+ */
+export const useCustomCompareMemo = <Factory extends () => unknown, Deps extends DependencyList>(
+  factory: Factory,
+  deps: Deps,
+  depsAreEqual: DepsAreEqual<Deps>
```
xobotyi, line 18: Better name this argument as `comparator` and reuse `IDependenciesComparator` type from `useCustomCompareEffect`.

Dont mind `I` prefix - all prefixes will be removed in close future.

src/useCustomCompareMemo/useCustomCompareMemo.ts
```diff
@@ -0,0 +1,33 @@
...
+) => {
+  const savedDeps = useRef<Deps>(deps);
+  const canCall = useRef(false);
```
xobotyi, line 21: there is no need to invoke whole hook that

```ts

const dependencies = useRef<Deps>();

if (dependencies.current === undefined || !comparator(dependencies.current, deps)) {
  dependencies.current = deps;
}

return useMemo(factory, dependencies.current)

```

xobotyi (changes requested):

xobotyi (approved):

Issue #951, opened 2022-10-02
useMeasure hook give wrong  values  : x,y,top,left are always 0


MarwenLabidi:
**simple example :** https://codesandbox.io/s/trusting-bush-4nhxe3?file=/src/App.js
**What is the expected behavior?**
I expect the result of useMeasure  top left x and y to match the ones from the DOM.

JoeDuncko:
@MarwenLabidi your `app.js` seems to be missing the useMeasure. This is all I see:

```
import "./styles.css";

export default function App() {
  return (
    <div className="App">
      <h1>Hello CodeSandbox</h1>
      <h2>Start editing to see some magic happen!</h2>
    </div>
  );
}

```

MarwenLabidi:
@JoeDuncko done , sorry for that

JoeDuncko:
Got a chance to look at this. It looks like our [example](https://react-hookz.github.io/web/?path=/docs/sensor-usemeasure--example) also doesn't have an example where the top, left, x, and y show anything but 0. I also can't force it to do anything but be 0.

I haven't played with ResizeObserver enough to know if this is expected behavior or not. Will research and find out. Or someone can chime in :P .

ArttuOll:
I can also take a look at this today. Will report here, if I find out anything.

ArttuOll:
When we use this hook, we are giving a `ref` to the element we want to measure. We can access this element's bounding rect by calling `.getBoundingClientRect()` on it. This rect contains the measures we are interested in here.

When I access the example's measured element using the Chrome devtools, I see very different values for `x` and `y` than 0, which seems to confirm, that there indeed is a bug here. I also get the differing measurements when I log the measurements of the `elementRef` on line 19 of the hook. The `x` and `y` properties are supposed to be relative to the viewport, which they clearly are not at the moment.

Continuing to investigate.

ArttuOll:
>  I also can't force it to do anything but be 0.

I found that the coordinates can be changed by adding padding to the measured element.

I explain the reason in my draft PR #954 : 
> The contentRect property's x and y coordinates, as well as the top and left properties, are defined by the element's top and left padding, as explained in the [spec](https://w3c.github.io/csswg-drafts/resize-observer-1/#content-rect). This does not appear to be the case for the rect returned by `getBoundingClientRect`.

lensbart:
The same issue exists for `react-use`. [This PR](https://github.com/streamich/react-use/pull/2031) might be helpful.

[...]

PR #954, opened 2022-10-06
fix(useMeasure): Fix x and y coordinates always being 0

ArttuOll:

Currently, when the rectangle in the [example](https://react-hookz.github.io/web/?path=/docs/sensor-usemeasure--example
) is resized, the `x` and `y` coordinates (as well as the `top` and `left` properties) of the measurement produced by `useMeasure` are always 0. These coordinates are supposed to represent the location of the measured element's [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly) relative to the viewport, which they clearly do not, at the moment.

Steps to reproduce:
1. Go to the [example](https://react-hookz.github.io/web/?path=/docs/sensor-usemeasure--example
) of the hook.
2. Resize the red rectangle.
3. Observe, that the `x` and `y` coordinates are always 0.

To affect the `x` and `y` coordinates at all, we must add top and left padding to the measured element.

The expected behavior is that the `x` and `y` coordinates would represent the actual coordinates of the element's DOMRect. In the example, the actual coordinates can be inspected by selecting the element by using the Chrome dev tools and typing `$0.getBoundingClientRect()` to the console.

Instead of getting the element's measures from the `contentRect` property of the `ResizeObserverEntry`, we get them through its `target` element using `getBoundingClientRect`.

**The `contentRect` property's x and y coordinates, as well as the top and left properties are defined by the element's top and left padding, as explained in the [spec](https://w3c.github.io/csswg-drafts/resize-observer-1/#content-rect). This [is not](https://drafts.fxtf.org/geometry-1/#domrect) the case for the `rect` returned by `getBoundingClientRect`. The difference is that `contentRect` is a content rect and `getBoundingClientRect` returns a DOMRect.**

## Checklist

  - #951 


ArttuOll:
The only problem I find with this solution is that for some reason, the first `x` and `y` coordinates of the measurements (after them being `undefined` on initial render) we get from the hook seem to be wrong. The first coordinates seem to be the coordinates of the _parent_ element's rect. When the rectangle in the [example](https://react-hookz.github.io/web/?path=/docs/sensor-usemeasure--example) is first resized, the coordinates become correct. 

Also, the initial values of the `x` and `y` coordinates seem to vary slightly when repeatedly refreshing the page and observing their values. I suspect it might have something to do with this:

From [MDN on getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)
> The amount of scrolling that has been done of the viewport area (or any other scrollable element) is taken into account when computing the bounding rectangle. This means that the rectangle's boundary edges (top, right, bottom, left) change their values every time the scrolling position changes (because their values are relative to the viewport and not absolute).

ArttuOll:
> The only problem I find with this solution is that for some reason, the first x and y coordinates of the measurements (after them being undefined on initial render) we get from the hook seem to be wrong. When the rectangle in the [example](https://react-hookz.github.io/web/?path=/docs/sensor-usemeasure--example) is first resized, the coordinates become correct.

Ok, now I see what is going on here. The reason why this happens is that the example has dynamic content: the printing of the measurements above the red rectangle. When I remove it and observe the measurements by logging them, the measurements are always the same as what the browser tells. What happens is that the initial value of the `y` coordinate and the `top` property are measured before the printing of the measurements happens. When the printing appears above the red rectangle, the `y` coordinate is not updated, since it gets updated only when the rectangle is resized. This might or might not be problematic.

JoeDuncko:
I've been thinking about this a bit, and I am concerned that `useMeasure` exposes `x` `y` despite only updating those values when the measurements of the component actually change. That seems to be a potential footgun if someone wants to rely on those values.

Thoughts?

To recreate in storybook: Change the size of the red square using the handle, scroll, then change the size of the red square again - the `y` position will only update when the square's measurements are updated.

ArttuOll:
> I am concerned that useMeasure exposes x y despite only updating those values when the measurements of the component actually change. That seems to be a potential footgun if someone wants to rely on those values.

Definitely agree. Now that I think of it, I feel like this hook should only expose `width` and `height`, the measures of the element, because it's called "useMeasure". There probably should be another hook that would expose the coordinates of an element and update them whenever scrolling happens. It could be called something like "usePosition" or "useCoordinates" ("useLocation" would be bad, because it's used by routing libraries).

ArttuOll:
@JoeDuncko Should we merge this anyway and open an issue about the point you raised above? This PR fixes a bug the users have. The point you raised relates to another bug, or unexpected behavior, that this hook has and maybe a separate issue would be in place?

JoeDuncko (approved):
LGTM

JoeDuncko:
> @JoeDuncko Should we merge this anyway and open an issue about the point you raised above? This PR fixes a bug the users have. The point you raised relates to another bug, or unexpected behavior, that this hook has and maybe a separate issue would be in place?

That sounds like a plan to me

JoeDuncko:
closed #951

xobotyi:
The only thing that disturbs me - `getBoundingClientRect` which causes DOM reflow. therefore each component that useess this hok will cause reflow on every change - performance wise it is a bad solution.

ArttuOll:
> The only thing that disturbs me - `getBoundingClientRect` which causes DOM reflow. therefore each component that useess this hok will cause reflow on every change - performance wise it is a bad solution.

Do you have any suggestions on how to proceed?

We could also keep using the `contentRect` property and just expose `width` and `height`. This would address both your and Joe's concerns.

xobotyi:
It is what im thinking of right now. Have to dig a bit more later this evening.

xobotyi:
Also there is a thing that entry contains several measurments that some might find useful, im talking about `[borderBoxSize](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/borderBoxSize)`, [contentBoxSize](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentBoxSize) and [devicePixelContentBoxSize](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/devicePixelContentBoxSize).
But, for that case, i think, devs are able to use resize-observer hook in case they need more infos.

xobotyi:
@ArttuOll so lets stick with plan to leave only width and height.
dont forget to change state only if any of dimesions changed

xobotyi:
Somehow i've missed your comment =)
Thanks @ArttuOll

PR #957, opened 2022-10-07
Add an optional argument to useIntervalEffect to synchronise updates

Rycochet:
Sadly not got time to do the documentation and tests for this in the next couple of weeks at least - but if anyone wants to add those bits to get this to a mergeable point then go for it!

This adds an optional synchronisation to useIntervalEffect - in other words, anything that uses the option and has the same update time will be called in the same update cycle. For instance allowing for multiple clocks on screen that all update in the same tick etc.

It uses two Maps, one simply stores the `setInterval` result so it can be removed when unused, and the second stores an array of callbacks. Whenever a callback is added or removed the array is mutated (rather than replaced - GC reasons mainly), and the callbacks themselves are wrapped in try/catch so they don't break anything. The dependencies list adds the flag coerced into a boolean - don't care what it actually is, but `undefined == false` and hooks use `===` so it saves processing on that edge case of `undefined` -> `false`.

### DIFF
--- a/src/useIntervalEffect/useIntervalEffect.ts
+++ b/src/useIntervalEffect/useIntervalEffect.ts
@@ -2,14 +2,47 @@ import { useEffect } from 'react';
 import { useSyncedRef } from '../useSyncedRef/useSyncedRef';
 
 /**
+ * A Map of delay -> `callSynchronized` intervals for removal when unused.
+ */
+const intervalMap = new Map<number, ReturnType<typeof setInterval>>();
+
+/**
+ * A Map of delay -> callbacks array.
+ *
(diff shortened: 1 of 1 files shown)


xobotyi:
It should be separate hook

Rycochet:
> It should be separate hook

Feel free to do anything with this code - it'll be a while before I can do anything about it - but not sure I agree on that, as an optional argument allows someone to choose whether it's sychronised or not without having to do some very non-hook code :-)

xobotyi:
It adds unncessary complexity to simple hook, in case somene required harmonized timeouts - it must be a separate hook. closing this PR - will do a separate hook when and if will have time to work on it.

Issue #941, opened 2022-09-19
cjs useCookieValue will include whole react-hookz rather than just one hooks


vip30:
### What is the current behavior?

Included whole lib in the bundle 

### Steps to Reproduce

1. one of the file using `import { useCookieValue } from "@react-hookz/web/cjs/useCookieValue/useCookieValue"`
2. use `webpack-bundle-analyzer` to view the bundle

### What is the expected behavior?

It should include useCookieValue only rather than whole bundle

I believe the potential cause is because we import the other hooks in this [syntax](https://github.com/react-hookz/web/blob/master/src/useCookieValue/useCookieValue.ts#L4)
`import { useFirstMountState, useMountEffect, useSafeState, useSyncedRef } from '..';`

PR #958, opened 2022-10-07
style: removes circular imports from the library

### LINKED ISSUE #905: Import structure is being detected as circular dependencies
### Prior Issues

Are there any existing issues or PRs that relate to this problem? If so, link them here.

Hooks all import dependencies from `../` (or the `index.js` file) which also exports all the hooks. This is being detected as a circular dependency in our build system.

Instead of importing from `../` the hooks should import from the dependant hook file directly

For example:
`import { useRafCallback } from '..';` becomes `import { useRafCallback } from '../useRafCallback/useRafCallback';`

### Environment Details

- _`@react-hookz/web` version: `15.0.1`
- _`react` version: `17.0.1` 
- _`react-dom` version: `17.0.1`
- _`typescript` version: `4.5.2`
- _Did this work in previous versions? unknown but probably not

wesgro:
This library is being detected as having circular imports which can prevent its use in some projects. This changes all the `import {<thing>} from '..'` lines to instead import directly from the files that have the required dependencies. This also moves some types to their own file to remove circular imports from the `utils` directory. Otherwise this change makes no functional change to any hooks.

Will close #905

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -46,8 +46,8 @@ If you are contributing for the first time - we would recommend reading
    - In case hook is stateful and exposes `setState` method, or is has async callbacks (that can
      theoretically be resolved after component unmount), it should use `useSafeState` instead of
      `useState`.
-   - In case of hooks reuse, import them as `import { useSafeState } from '..';` instead of
-     `import { useSafeState } from '../useSafeState/useSafeState';`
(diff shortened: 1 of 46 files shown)


ArttuOll:
Also possibly related #941.

xobotyi:
fix: #941 
fix: #905

Issue #451, opened 2021-11-12
useSessionStorageValue same key, different value


JaminFarr:
### Prior Issues

I cannot see any related issues.

### What is the current behavior?

Different instances of `useSessionStorageValue` with the same key that are mounted at the same can get out of sync when setValue is use on first mount within a useEffect. Only the component making the setValue call, and components before it, get the updated value. Components after do not update and have a stale value.

### Steps to Reproduce

Using `setValue` in a `useEffect(..., [])`.

```js
const useStoredValue = () =>
  useSessionStorageValue("example-key", "Default Value");

const UpdateOnMount = () => {
  const [, setValue] = useStoredValue();

  useEffect(() => {
    setValue("Update Value");
  }, [setValue]);

  return null;
};

const DisplayValue = () => {
  const [value] = useStoredValue();

  return <p>Stored value is "{value}"</p>;
}

const App = () => (
  <div>
    <DisplayValue /> 
    <UpdateOnMount />
    <DisplayValue />
  </div>
)
```
codesandbox: https://codesandbox.io/s/eager-gould-ter7m?file=/src/App.js

The code above on first load displays:
```
Stored value is "Update Value"

Stored value is "Default Value"
```

Another codesandbox with random values and reset buttons: https://codesandbox.io/s/fervent-stitch-zqekk

There's a few work arounds:
 * Changing the component order
 * Using `initializeWithStorageValue: false`
 * Putting setValue in a setTimeout
```js
  useEffect(() => {
    setTimeout(() => {
      setValue("Update Value");
     }, 0);
  }, [setValue]);
``` 

### What is the expected behavior?

All instances of `useSessionStorageValue` with the same key should return the same value.

### Environment Details

- _`@react-hookz/web` version:_ 12.0.0
- _`react` version:_ 17.0.2 
- _`react-dom` version:_ 17.0.2 
- _`typescript` version:_ N/A
- _OS:_ macOS BigSur
- _Browser:_ Chrome 95 / Firefox / Safari
- _Did this work in previous versions?_ No know

xobotyi:
Well, nope, didnt worked out. Ill dig more later.

ArttuOll:
I'm investigating this also. Will report here if I find anything out.

ArttuOll:
I can confirm that:
- This affects both `useLocalStorage` and `useSessionStorage` (which is not surprising, since both use `useStorage` under the hood).
- This bug has been present since `useLocalStorage` and `useSessionStorage` were first added to this library.

PR #959, opened 2022-10-08
fix(useStorageValue): Allow setting of state even if the component has not mounted

ArttuOll:

The current behavior has been described in #451. In short, using the `setValue` function provided by `useSessionStorageValue` and `useLocalStorageValue` on first mount could cause other instances of those hooks to get out of sync.

Every instance of the `useSessionStorageValue` and `useLocalStorageValue` hooks should return the same value at all times.

By using `useState` to store the value of the tracked key instead of `useSafeState`.

In the sandbox below, I have modified the `useSessionValue` hook in the above way. The example is the one provided in #451 , but this time the reported issue is not present.

[Edit unruffled-dijkstra-kqnbjy](https://codesandbox.io/s/unruffled-dijkstra-kqnbjy?fontsize=14&hidenavigation=1&theme=dark)

Why does this work, you might ask? I do not know. I will have to digest this a bit more. 

I arrived at this solution from the simple notion that the `state` variable is what gets out of sync. I then started going through everything that affected `state`. I went through many things, but finally I noticed that `useSafeState` also "affects" `state` (in a sense that it is not the standard `useState` that I picture in my mind when I think about React state). I then tried switching it to
a traditional `useState`, which seemed to solve the issue.

### Checklist

- [X] Have you read [contribution guideline](https://github.com/react-hookz/CONTRIBUTING.md)?
  -  #451 

### DIFF
--- a/src/useStorageValue/useStorageValue.ts
+++ b/src/useStorageValue/useStorageValue.ts
@@ -1,11 +1,10 @@
 /* eslint-disable @typescript-eslint/no-use-before-define,no-use-before-define */
-import { useCallback } from 'react';
+import { useCallback, useState } from 'react';
 import { useConditionalEffect } from '../useConditionalEffect/useConditionalEffect';
 import { useFirstMountState } from '../useFirstMountState/useFirstMountState';
 import { useIsomorphicLayoutEffect } from '../useIsomorphicLayoutEffect/useIsomorphicLayoutEffect';
 import { useMountEffect } from '../useMountEffect/useMountEffect';
 import { usePrevious } from '../usePrevious/usePrevious';
-import { useSafeState } from '../useSafeState/useSafeState';
 import { useSyncedRef } from '../useSyncedRef/useSyncedRef';
 import { useUpdateEffect } from '../useUpdateEffect/useUpdateEffect';
 import { NextState, resolveHookState } from '../util/resolveHookState';
@@ -157,7 +156,7 @@ export function useStorageValue<T>(
   });
 
   const isFirstMount = useFirstMountState();
-  const [state, setState] = useSafeState<T | null | undefined>(
+  const [state, setState] = useState<T | null | undefined>(
     initializeWithStorageValue && isFirstMount ? (methods.current.fetchVal() as T) : undefined
   );
   const prevState = usePrevious(state);


ArttuOll:
> Why does this work, you might ask? I do not know. I will have to digest this a bit more.
Maybe what happens is something like this (referring to the components in my codesandbox):
1. `ValueDisplay` mounts and displays the current value in the storage.
2. `Provider` mounts, changing the value in the storage.
3. `ValueDisplay` has already mounted, so `useSafeState` allows its state to update.
4. The components after `Provider` have not yet mounted and thus `useSafeState` does not allow their states to be updated. When they mount, they display the value which their `useSessionStorage` hooks had when they were initialized.

Just a quick thought.

Ps. that codecov/patch action certainly seems to take its time 😄

xobotyi:
Well, okay, the reason - i've finally got it =) `StrictMode` + `useFirstMountState` - sandbox is using strict mode, if to disable it - issue not presented.

What we know about strict mode - effects are triggered twice, `useFirstMountState` unable to work wtithin strict mode, since it is based off `useRef`.

ArttuOll:
Oh, nice! I should have thought about `StrictMode` 😆 It's so easy to forget, but should be taken into account when debugging...

ArttuOll:
But what should be done about this? This bug is very distracting, even if it only exists in `StrictMode`, ie. during development.

xobotyi:
I'm trying to make a solution at the moment, but i see literally no chance of making `useIsMounted` survivable to doubled effects

xobotyi:
Ofc we can make go with solution you proposed, but it is only fix of single symptom, not its source.

ArttuOll:
Well, this relates to another thing I had in mind earlier today. Why are we using `useSafeState` in the first place?

According to the [React team](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html) it is an anti-pattern to check whether the component is mounted or not before setting state. They argue that any state setting functions should be cancelled before the component unmounts instead. They go more detail into this in the blog post. [This](https://medium.com/doctolib/react-stop-checking-if-your-component-is-mounted-3bb2568a4934) article talks about the same topic, but is a bit more fresh.

ArttuOll:
To expand on my previous comment:

From the perspective of a hook library, it essentially should be the user of the stateful hook's responsibility to check that the state is not set after the component has unmounted, not the library's.

xobotyi:
Mostly `useSafeState` used within hooks that are making use of any WEB API or has event handlers, where any event may occur during component unmounts.

ArttuOll:
> Mostly useSafeState used within hooks that are making use of any WEB API or has event handlers, where any event may occur during component unmounts.

Indeed. And the articles I linked both argue that in such situations, it would be correct to cancel the state settings functions (maybe an async event handler or a `useEffect` making a web request) when the component unmounts, rather than checking for the mounted state before settings state.

As pointed in the second of the articles I referenced, `useSafeState` merely hides the _Can’t perform a React state update on an unmounted component_ warning. React already does the `isMounted` check before updating state and gives that warning, if a component is not mounted when a state update is attempted. Doing any `isMounted` checks application-side merely move the check from React internals to the application, hiding the warning.

xobotyi:
it is impossible to cancel async chain midair, in case processing went further direct handlers - you're stuck
im taling about
```ts
Promise.
then(()=>{
 await Promise2ThatTakesLotsOfTime()
 // unmount happens here
})
```

ArttuOll:
> Mostly useSafeState used within hooks that are making use of any WEB API or has event handlers, where any event may occur during component unmounts.

And to respond in the context of this library, `useSafeState` is also heavily used internally, which appears unnecessary to me, and it also seems to have caused the bug this PR solves (in a way).

ArttuOll:
> it is impossible to cancel async chain midair, in case processing went further direct handlers - you're stuck im taling about
> 
> ```ts
> Promise.
> then(()=>{
>  Promise2ThatTakesLotsOfTime().then(setState)
>  // unmount happens here
> })
> ```

Isn't this possible using the `AbortController` API, maybe in combination with `useEffect`?

From the second article
```typescript
useEffect(() => {
  const abortController = new AbortController()   // creating an AbortController
  fetch(url, { signal: abortController.signal })  // passing the signal to the query
    .then(data => {
      setState(data)                              // if everything went well, set the state
    })
    .catch(error => {
      if (error.name === 'AbortError') return     // if the query has been aborted, do nothing
      throw error
    })
  
  return () => {
    abortController.abort()                       // stop the query by aborting on the AbortController on unmount
  }
}, [])
```

In any case, in the article the React team seems to suggest that such state setting functions should be made cancellable.

xobotyi:
TBC im sitting with a thought of dropping lots of `useSafeState` usages for a month or so.
React18 anyway dropped the message about setting state on unmounted component.

ArttuOll:
I'm glad to hear that! Let me know, if I can help in refactoring away the `useSafeState`'s.

xobotyi:
As said - yes, abort controller kinda solves situation, but it is not widely supported.
> and frankly speaking its usage quite ugly, only try\catch blocks are uglier.

ArttuOll:
> As said - yes, abort controller kinda solves situation, but it is not widely supported.
> 
> > and frankly speaking its usage quite ugly, only try\catch blocks are uglier.

`AbortController` not having wide support if a fair point. Plain promises surely can be cancelled somehow by (ab)using `Promise.reject` but I don't even want to imagine how tedious that would be. That would be appear to be necessary, however.

Also, using `AbortController` can get pretty ugly, but `useAsyncAbortable` makes it a lot prettier 😉 .

xobotyi:
welp, everything ended up with complete `useStorageValue` rework :D

ArttuOll:
That is probably good, the code felt quite difficult before 😄

I got kind of inspired by our conversation here. What do you think about this:

1. I open an issue with a complete write up of our discussion here so that everybody can easily see it and take part.
2. I make a PR with all the internal usages of `useSafeState` removed.

ArttuOll:
Also, I noticed that you used this PR's solution for the bug as part of #960 . Do you mind adding the hacktoberfest-accepted label to this PR?

xobotyi:
@ArttuOll, are you planning to continue work on react-hookz or is it just hacktoberfest related?

ArttuOll:
> There should be another label for that? i thought just hascktoberfest would be fine

Yeah, if this PR is not merged, but the code is accepted, that's how they know to count this to my score.

>Will you be working on this after October

Yes, so no worries that I would just drop everything that I'm working on when this month ends 😁 perhaps not as actively

xobotyi:
@ArttuOll what do you think joining team and becoming maintainer?

ArttuOll:
Well, I'm pretty sure I would be able to find time for that and that would surely be interesting, so count me in!

PR #960, opened 2022-10-11
Rework `useStorageValue` to more simple and robust variant

### LINKED ISSUE #451: useSessionStorageValue same key, different value

xobotyi:
New hook has less options, now it is impossible to make hook isolated, all hooks listening same key are synchronised.

Also hook return signature changed to object instead of array, as it provides more obvious API.

close #959
fix #451


xobotyi:
@ArttuOll could you please finish this one?

I think only docs left for now, the code itself and tests are already done.

ArttuOll:
Sure! I'll take a look over the next few days.

xobotyi:
@JoeDuncko @ArttuOll please review PR as a whole.

src/useStorageValue/useStorageValue.ts
```diff
@@ -1,336 +1,267 @@
...
-    on(window, 'storage', storageHandler, { passive: true });
-
-    return () => {
-      off(window, 'storage', storageHandler);
-    };
+  useEffect(() => {
+    if (!options?.initializeWithValue) {
+      stateActions.current.fetch();
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
ArttuOll, line 178: I noticed that here we don't actually need to disable this lint rule since `stateActions` is a ref (ESLint knows that refs don't have to be included in the dependency array, but it does not know that `stateActions` is a ref because it comes from our own custom hook) and `options?.initializeWithValue` does not really change unless the developer makes it change by controlling it with a prop coming from the parent (which might be a rare but valid usecase that we are blocking currently).

src/useStorageValue/useStorageValue.ts
```diff
@@ -1,336 +1,267 @@
...
-const storageKeysUsed = new Map<Storage, Map<string, Set<CallableFunction>>>();
+  return useMemo(
+    () => ({
+      value: state as UseStorageValueValue<Type, Default, Initialize>,
+      set: actions.current.set,
+      remove: actions.current.delete,
+      fetch: actions.current.fetch,
+    }),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [state]
```
ArttuOll, line 226: The same thing here with the lint rule. This dependency array could also include `actions`, because it too is a ref.

(I always pay attention to disabling of this rule, because I've seen it abused too many times at work, causing mysterious and hard-to-debug behavior. Not saying that's the case here.)

```suggestion
    [state, actions]

ArttuOll (commented):
The hook seems to work as expected.

One thing I noticed was that the code example does not show the usage of the hook itself; it just shows the rendered HTML. I could not figure out how to make it show the hook though.

I commented on a couple spots where the `exhaustive-deps` lint rule seemed to be unnecessarily disabled, which I personally dislike a lot, but it does not really affect the functionality of the hook here.

src/useStorageValue/useStorageValue.ts
xobotyi: It should only be used once on first render, therefore there is no legit scenario this hook should be retriggered on option change.

src/useStorageValue/useStorageValue.ts
xobotyi: As for here it is simply useless to add it to dependencies as actions never changes, therefore it would be useless comparison on every render.

xobotyi:
There is barely possible example of `useStorageValue` hook, also it is not exported at all.

As for `useSessionStorage/LocalStorage` - they're pretty functional, based on inputs

src/useStorageValue/useStorageValue.ts
ArttuOll: Yeah, it does not matter - personally I woud rather add it than add the comment, but does not really matter

ArttuOll:
> There is barely possible example of `useStorageValue` hook, also it is not exported at all.
> 
> As for `useSessionStorage/LocalStorage` - they're pretty functional, based on inputs, showing that hooks are synced across the page and, in case of LocalStoage - across tabs

Sorry, I think I was a bit unclear. I meant that the examples of `useLocalStorageValue` and `useSessionStorageValue` do not show the usage of the hook in their code examples (when you press the "show code" button in Storybook).

src/useStorageValue/useStorageValue.ts
ArttuOll: Perhaps it isn't. It certainly would be odd and rare 🤔

xobotyi:
Have no idea how to make storybook to better handle code preview =(
Also have an idea of migrating to docosaurus since storybook caused lots of issues recently, but it is another discussion =)

ArttuOll:
We have other examples also where this is a problem. I was thinking of going over the docs some time and doing little improvements like this. I'll worry about this then.

ArttuOll (approved):

PR #963, opened 2022-10-12
feat: remove cjs support and raise language level to ESNext

### LINKED ISSUE #962: Raise the cjs and esm target to es6
## New Features

The ESM and CJS targets use es6 instead of es5

All evergreen browsers fully support the ES6 spec (Even Safari as of Safari 10)

Generating ES5 compliant code introduces unnecessary feature polyfills that can bloat bundles and possibly create warnings in rollup (no top level `this` for example).

### What is the new or updated feature that you are suggesting?

Change the `target` properties for the CJS and ESM builds to use es6 instead of es5

### Why should this feature be included?

Since this library supports React 18 and React 18 no longer supports IE 11 this seems like a good time to raise the floor of the non-ESNext generated code.

If users of this library need to support IE 11 still they can always include this library in their babel config to transpile it down to whatever ES version they need to support.

wesgro:
BREAKING CHANGE

All evergreen browsers fully support the ES6 spec (Even Safari as of Safari 10)

Generating ES5 compliant code introduces unnecessary language polyfills that can bloat bundles and possibly create warnings in rollup (no top level this for example).

Since this library supports React 18 and React 18 no longer supports IE 11 this seems like a good time to raise the floor of the non-ESNext generated code.

If users of this library need to support IE 11 still they can always include this library in their babel config to transpile it down to whatever ES version they need to support.

Fixes #962


ArttuOll:
I find your reasoning to be sound. ES6 indeed [is massively supported](https://caniuse.com/es6).

The only thing that worries me is this statement from the [browserlist](https://github.com/browserslist/browserslist) repo:
> Don’t remove browsers just because you don’t know them. Opera Mini has 100 million users in Africa and it is more popular in the global market than Microsoft Edge.

I was able to confirm the 100 million users from the Opera Mini site.

It happens to be the case that Opera Mini [does not support ES6](https://caniuse.com/es6). If this change was made, all the developers developing for those 100 million users in Africa would have to transpile this library.

This is all just food for thought. In my opinion, this change would be good. Opera Mini should catch up and add support for ES6. I also checked a bunch of popular React libraries (such as react-query and NextJS) and their TypeScript targets also seem to be either es6 or esnext.

What do you think @xobotyi?

xobotyi:
I dont have a decision tbh.
As for my projects i transpile node_modules anyway, since i have to maintain compatibility with console browsers (which is complete mess tbh).

Im not able to imagine how many users pipelines that change would break.
I'd consider hanging it for a while, untill we'll clear all current issues and then make a major version with lang-level upgrade

wesgro:
> Im not able to imagine how many users pipelines that change would break.
Would you be open to only the `esm` path using `es6`? ESM would only be in use by browsers that have es6 support already. the CJS path could remain in es5 form.

JoeDuncko:
> > Im not able to imagine how many users pipelines that change would break.
> 
> Would you be open to only the `esm` path using `es6`? ESM would only be in use by browsers that have es6 support already. the `cjs` path could remain in es5 form.

This sounds like a reasonable compromise to me. Anyone else?

ArttuOll:
> Would you be open to only the esm path using es6? ESM would only be in use by browsers that have es6 support already. the cjs path could remain in es5 form.

Sounds like a really reasonable compromise to me as well.

xobotyi:
I've reduced esnext version to esmodules.
This patch will go as major version as it breaks backwards compatibility. (but is highly doubdt that someone used esnext version =))

wesgro:
I see a few breaking changes have gone through, is there a plan for this to get in soon?

xobotyi:
Yes, as soon as we figure current backlog in terms of backlog

xobotyi:
*current backlog before this pr

ArttuOll:
> Yes, as soon as we figure current backlog

@xobotyi On our issues list there's currently pretty much nothing in addition to this that we would need to act on. There's open PR's but I don't think any of them are ready to be merged in a while, except for the documentation PRs. Would now be the time to tackle this PR?

xobotyi:
yup - i've planned to handle it on tuesday

xobotyi:
Im thinking of making one more step further and leave only one version - esm, with esnext langlevel.

A bit harsh move, but we have make it anyway as we will have to move to modules once infrastructure will be ready. All major bundles handles module dependencies perfectly.

I'll update this pr now, and leave it here for a couple of days. planned date of merge is 4th Jan.

xobotyi:
the only downsite - had to temporarily disable built bundle testing die to issue with jest: https://github.com/facebook/jest/issues/13715

wesgro:
@xobotyi Would it be possible to retain `cjs` support? Although many modern bundlers have support there are probably a tonne of legacy bundlers still in use that would need a CJS path (I say greedily, as I have to deal with one of them 😜 )

xobotyi:
Which one? 🤔

wesgro:
I believe we still have something using webpack 3 and requires us to ship CJS or they can't use our library. They're a different team from us so them changing their build for our teams' needs probably isn't something that will get on their road map any time soon.

Not the end of the world since I could opt to bundle this code when we ship CJS, but not completely ideal either.

xobotyi:
But you have to transpile `node_modules` anyway, since many tools coming esm only🤔

As for me - I'd cut everything in one move, since it will ease further support and future moving to `type: module`

wesgro:
Sounds good 👍 I don’t want to be in the way of progress

xobotyi:
@wesgro in closest release CJS version will be returned🙃

wesgro:
Moving fast and breaking things!

PR #973, opened 2022-10-21
fix: migration guide

lensbart:
## What is the problem?

`useWindowSize` was listed as not implemented in the migration guide

Migration guide now links to `useWindowSize` doc


ArttuOll (approved):
`useWindowSize` indeed is implemented.

PR #974, opened 2022-10-22
`useHover`

lensbart:
Feel free to add tests

## What new hook does?
https://github.com/streamich/react-use/blob/master/src/useHover.ts
https://streamich.github.io/react-use/?path=/story/sensors-usehover--docs

## Checklist

- N/A Does the code have comments in hard-to-understand areas?
- [ ] Have the tests been added to cover new hook?

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,8 @@ Coming from `react-use`? Check out our
 
 - #### Sensor
 
+  - [**`useHover`**](https://react-hookz.github.io/web/?path=/docs/sensor-usehover--example)
+    — Track if an element is being hovered by a mouse.
   - [**`useIntersectionObserver`**](https://react-hookz.github.io/web/?path=/docs/sensor-useintersectionobserver--example)
     — Observe changes in the intersection of a target element with an ancestor element or with the
     viewport.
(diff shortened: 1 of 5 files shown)


JoeDuncko:
Thanks for the PR! We'll definitely need to add tests to this before merging, but this is a great start.

xobotyi:
I'm not completely sure if we should finish implementing this hook or dump it.

Also i really dont like the amount of 'machinery' used to simply add 2 event handlers - i dont understand why you even need to clone element to add handlers🤔

Most of cases usually covered by ui-kit libs, others can easily be implemented with literally 3 lines of code in mode versatile way like following.

```
const [state, setState] = useState<boolean>(false);
useEventListener(tgtRef, 'mouseenter', ()=>{setState(true)})
useEventListener(tgtRef, 'mouseleave', ()=>{setState(false)})
```

@JoeDuncko @ArttuOll WDYT?
i vote dropping this hook with adding note within migration guide.

ArttuOll:
Yeah let's drop it. If I recall correctly, I looked into this and found that this hook would also be difficult to test. In fact, it is untested in `react-use`.

Issue #1002, opened 2022-11-03
useLocalStorageValue yields undefined on firstRender if initializeWithValue not set


Evalon:
### What is the current behavior?

`useLocalStorageValue` yields undefined on first render if `initializeWithValue` not set to true and on second render yield actual value from localStorage.

https://codesandbox.io/s/wispy-forest-qmrlzn?file=/src/App.tsx

### What is the expected behavior?

`initializeWithValue: true` should be default behavior based on documentation and previous behaviour.

### Environment Details

- _`@react-hookz/web` version: 17.0
- _`react` version: 18.2 
- _`react-dom` version: 18.2
- _`typescript` version: 4.8.4
- _OS: macos
- _Browser: chrome
- _Did this work in previous versions? yes

ArttuOll:
Thanks for the bug report!

I managed to repeat this. Indeed, it seems that the default value of `initializeWithValue` is `false` even though in the code (and in the docs) it definitely looks like it should be `true`.

I'm looking into this.

PR #1003, opened 2022-11-04
Fix `initializeWithValue` default not respected, fix docs and tests

ArttuOll:

`initializeWithValue` default value is not respected. It is always read from user-given `options`, meaning that it is `undefined` (effectively `false`), if the user did not explicitly set it.

Reproduction steps are given in #1002 . It looks like `initializeWithValue` is `false` by default (which is what the bug report is about), but actually it's just that the default value of `true` is not respected.

Additionally, I noticed that the default value for `initializeWithValue` is set to `true`, which partly contradicts the documentation (in the top part of the docs of `useSessionStorageValue` and `useLocalStorageValue` `initializeWithValue` is said to be `false` by default, but in the description of the `options` argument it is said to be `true`) and the tests. After fixing `useStorageValue` to respect the default value of `initializeWithValue` it became necessary to set it to `false`, otherwise SSR tests would fail, as they should.

The `initializeWithValue` default should be respected and it should be `false` by default.

- Make sure the default value of `initializeWithValue` is used if it is not provided by the user.
- Set `initializeWithValue` to `false` by default.
  - Note, that this is not a breaking change, because previously it was already effectively `false`.

## Checklist

  - #1002 

### DIFF
--- a/src/useStorageValue/useStorageValue.ts
+++ b/src/useStorageValue/useStorageValue.ts
@@ -120,9 +120,9 @@ export interface UseStorageValueResult<
   fetch: () => void;
 }
 
-const DEFAULT_OPTIONS: UseStorageValueOptions<null, false> = {
+const DEFAULT_OPTIONS: UseStorageValueOptions<null, true> = {
   defaultValue: null,
-  initializeWithValue: false,
+  initializeWithValue: true,
 };
 
 export function useStorageValue<
--- a/src/useStorageValue/useStorageValue.ts
+++ b/src/useStorageValue/useStorageValue.ts
@@ -156,7 +156,9 @@ export function useStorageValue<
 
   const isFirstMount = useFirstMountState();
   const [state, setState] = useState<Type | null | undefined>(
-    options?.initializeWithValue && isFirstMount ? storageActions.current.fetch() : undefined
+    optionsRef.current?.initializeWithValue && isFirstMount
+      ? storageActions.current.fetch()
+      : undefined
   );
   const stateRef = useSyncedRef(state);
 
@@ -172,7 +174,7 @@ export function useStorageValue<
   }, [key]);
 
   useEffect(() => {
-    if (!options?.initializeWithValue) {
+    if (!optionsRef.current.initializeWithValue) {
       stateActions.current.fetch();
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
--- a/src/useStorageValue/useStorageValue.ts
+++ b/src/useStorageValue/useStorageValue.ts
@@ -120,9 +120,9 @@ export interface UseStorageValueResult<
   fetch: () => void;
 }
 
-const DEFAULT_OPTIONS: UseStorageValueOptions<null, true> = {
+const DEFAULT_OPTIONS: UseStorageValueOptions<null, false> = {
   defaultValue: null,
-  initializeWithValue: true,
+  initializeWithValue: false,
 };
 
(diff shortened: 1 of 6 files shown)


Evalon:
Hi!
Thank you for quick response and fix. I'm not sure about setting `initializeWithValue` to be false by default. In mind of developers `useLocalStorage` is the abstraction that works similar to `useState` but save value between user sessions in local storage. So if it yields undefined first and then yields actual value it breaks this abstraction.
Can we support this abstraction by default?

ArttuOll:
> Hi! Thank you for quick response and fix. I'm not sure about setting `initializeWithValue` to be false by default. In mind of developers `useLocalStorage` is the abstraction that works similar to `useState` but saves value between user sessions in local storage. So if it yields undefined first and then yields actual value it breaks this abstraction. Can we support this abstraction by default?

I understand your concern. However, supporting SSR is one of the key goals of this library. Settings `initializeWithValue` to false by default means SSR is supported out of the box. I'm interested to hear what my fellow maintainers have to say about this.

xobotyi:
That is the change from previous implementation, previously SSR users had to set this option to false, wheras this time it is set to `false` by default. I dont have complete decision for it - maybe we have to change it back to `true`

src/useStorageValue/useStorageValue.ts
```diff
@@ -156,7 +156,9 @@ export function useStorageValue<
 
   const isFirstMount = useFirstMountState();
   const [state, setState] = useState<Type | null | undefined>(
-    options?.initializeWithValue && isFirstMount ? storageActions.current.fetch() : undefined
+    optionsRef.current?.initializeWithValue && isFirstMount
```
xobotyi, line 159: There is no sence of using the ref value since it is only suitable during forst render, after re-render it wint take any effect -> no need to use ref.

src/useStorageValue/useStorageValue.ts
```diff
@@ -172,7 +174,7 @@ export function useStorageValue<
   }, [key]);
 
   useEffect(() => {
-    if (!options?.initializeWithValue) {
+    if (!optionsRef.current.initializeWithValue) {
```
xobotyi, line 177: Same as above

xobotyi:
So basically #1003 is about changed default value and docs not being fully actualized.

xobotyi:
@ArttuOll I'd loke you to fix the docs and make default value to be true. Also docs has to be actualized.

ArttuOll:
I see. I'll edit this that way then.

xobotyi:
yeah, sorry🙏 been distracted during design of reworked hook 🙈

Evalon:
Yeah, reflecting on it a little bit more. 
I also don't have a good decision for this. Defaults in context (like ssr=true)? Static initialization? Different hooks for CSR?
I fully understand your concerns. 
If `initializeWithValue` stays `false` then I need to make a wrapper around `useLocalStorageValue` and if it changes to `true` SSR-users need to make a wrapper like this. Because, it's inconvenient to remember to specify the same options everywhere in the case of CSR or SSR.

xobotyi:
or you can just `know` that you're using ssr ant set proper field. by now - none been complaining, since it is only hook that supports ssr\csr duality🙃

ArttuOll:
Btw, this becomes a breaking change now since we are setting `initializeWithValue` to true while it was previously effectively false.

xobotyi:
That was a 'bug' 🙃

PR #1004, opened 2022-11-04
Pr/remove hooks modifying head

ArttuOll:
## What is the problem?

In #386 it was decided that hooks modifying the contents of the `<head>` element should not be implemented and existing such hooks should be removed.

- Delete `useDocumentTitle` (are there any other that should be removed?)
- Update docs on `useDocumentTitle` and `useFavicon` with information about this decision and suggestion on using libraries such as `react-helmet`.
- My editor was screaming about typos and spelling errors in the migration guide, so I fixed them.

Closes #386


xobotyi:
LGTM, but! we have to fix useStorageValue first

xobotyi (approved):

ArttuOll:
One approving review and the CI is passing (coverage dropped merely because there's now less lines of code to test). Off to production we go!

Issue #1005, opened 2022-11-07
Function reference for useLocalStorageValue and useSessionStorageValue say that default value is a separate argument


ArttuOll:
## What docs page needs to be fixed?

- **Section**: Side-effect
- **Page**: useLocalStorageValue and useSessionStorageValue

## What is the problem?

The function references incorrectly say that `defaultValue` is the second argument, which is no longer true. `defaultValue` is now part of `options` which is the actual second argument.

PR #1010, opened 2022-11-08
docs(useLocalStorageValue & useSessionStorageValue): Fix function reference

ArttuOll:
## What is the problem?

The function references are the ones from the previous implementation.

Update them to the new implementation.

## Checklist

  - #1005


xobotyi:
Aah, only in docs. i see.

xobotyi (approved):

Issue #1000, opened 2022-11-01
Ensure that `useMediaQuery` has the "correct" initial value


Kyzegs:
### What is the new or updated feature that you are suggesting?

Ensure that `useMediaQuery` returns the matched value for the given query in the first render.

### Why should this feature be included?

In most cases, we do not want to load our animations on the initial page load. When we have a component with animations that relies on the `useMediaQuery` hook the following happens: The animation gets triggered because of the value from the hook changes. I've tried this with my animation implementation and libraries such as [Framer Motion](https://www.framer.com/motion/).

Having the correct initial value would fix this problem.

Kyzegs:
I'll try to have an environment with the issue up later today.

Kyzegs:
Here's the eEnvironment](https://stackblitz.com/edit/react-ts-fz7wbn?file=App.tsx) with the issue.

ArttuOll:
I looked into this and I understand your problem. I'm currently trying to figure out a solution.

kylemh:
I wanted to suggest reverting this and/or thinking about it more. I've just upgraded this package to the latest version and have to replace all my usages of `useMediaQuery` to include a second argument of `{ initializeWithValue: false }` in order to work in an SRR-based environment.

One suggestion might be to adjust [this line](https://github.com/react-hookz/web/pull/1020/files#diff-a4a82ae81629dbe9fe162c95cd2348f18617081fa3830992478adc9f02fa91fbR71):

`if (options?.initializeWithValue ?? true)` to be: `if (typeof window !== 'undefined' && (options?.initializeWithValue ?? true))`

I think this still appeases the original request without ruining the DX for people that use this hook with Next.js or Remix.

xobotyi:
Your code will cause hydration error, since SSR state and browser one will differ.

kylemh:
I see. I didn't think about strict mode... I'm not sure what the workaround is, but it really sucked to be punished by this change.

xobotyi:
Overall - i've missed it during review, anyway since all other hook causes hydration error on missconfig rather than SSR crash.

@ArttuOll please, add `isBrowser` guard to the code so SSR at least works.

ArttuOll:
No problem. I'll take a look later today.

ArttuOll:
@kylemh Just wanted to let you know that we're working on this feature #1023 , which would improve DX for SSR significantly. It's still a WIP though - all help appreciated.

PR #1020, opened 2022-11-20
fix(useMediaQuery): Return boolean describing media query match on first render

ArttuOll:

Described in #1000 .

The value returned from the hook should be a boolean describing the media query match on first render instead of `undefined`.

By initializing the state with the the boolean received from `matchMedia` and creating a `MediaQueryList` entry to the cache in the process.

[Here](https://codesandbox.io/s/jovial-yalow-rt5mn7?file=/src/App.tsx) is a CodeSandbox, which contains the example provided in #1000, but with the new implementation of `useMediaQuery`. Note that the animation is no longer triggered when the page loads.

**This is a breaking change, since the value returned on the first render is now different than before. (Returning `undefined` before was not a bug since it was documented)**

## Checklist

  - #1000 


xobotyi:
One problem - it will cause hydration error for SSR.

Initialisation should be flagged via hook parameters.

ArttuOll:
> One problem - it will cause hydration error for SSR.
> 
> Initialisation should be flagged via hook parameters.

Sounds doable. I'll implement this later.

ArttuOll:
Btw, could you elaborate on how this causes an hydration error? I wasn't aware of this.

xobotyi:
SSR will be first-rendered with undefined, client with some value - react will freak out on that.

xobotyi:
Also I have an idea to replace all that crap with `initializeWithValue` props in favor of context provider that will automatically tell all our hooks that we're in SSR mode without the need of passing extra arguments to hooks

ArttuOll:
> first render on server and in browser should match

Of course, makes sense! Thanks.

> Also I have an idea to replace all that crap with initializeWithValue props in favor of context provider that will automatically tell all our hooks that we're in SSR mode without the need of passing extra arguments to hooks

Interesting. Would the users then have to wrap their app in some sort of context provider?

xobotyi:
Yes, they should have, but in my eyes it is way more convenient that to be sure that you pass proper hook config to each hook that has isomorphic behavior.

ArttuOll:
@xobotyi How about something like I pushed in my last commit? I added `initializeWithValue` option and defaulted it to `true`, like with other hooks such as `useLocalStorageValue` etc.

src/useScreenOrientation/useScreenOrientation.ts
```diff
@@ -9,7 +9,7 @@ export type ScreenOrientation = 'portrait' | 'landscape';
  * hook uses CSS3 `orientation` media-query to check screen orientation.
  */
 export function useScreenOrientation(): ScreenOrientation | undefined {
-  const matches = useMediaQuery('(orientation: portrait)');
+  const matches = useMediaQuery('(orientation: portrait)', { initializeWithValue: false });
```
xobotyi, line 12: Maybe we should also add an argument to this hook to yield correct value on init

xobotyi (changes requested):

src/useScreenOrientation/useScreenOrientation.ts
ArttuOll: Good point. Did not even think about this hook too much 😄 I'll work on this tomorrow.

ArttuOll: Done.

PR #1045, opened 2022-12-09
Migrate `useQueue` from react-use

jpwallace22:
## What new hook does?
A react state hook implements simple first-in-first-out (FIFO) queue migrated over from [react-use](https://github.com/streamich/react-use)

## Checklist

  - _link issue here_

### DIFF
--- a/src/useQueue/__docs__/story.mdx
+++ b/src/useQueue/__docs__/story.mdx
@@ -6,7 +6,7 @@ import { ImportPath } from '../../__docs__/ImportPath';
 
 # useQueue
 
-A react state hook implements simple first-in first-out (FIFO) queue.
+A react state hook implements a simple first-in first-out (FIFO) queue.
 
 #### Example
 
--- a/src/useQueue/__docs__/story.mdx
+++ b/src/useQueue/__docs__/story.mdx
@@ -1,6 +1,6 @@
 import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
 import { Example } from './example.stories';
-import { ImportPath } from '../../storybookUtil/ImportPath';
+import { ImportPath } from '../../__docs__/ImportPath';
 
 <Meta title="State/useQueue" component={Example} />
 
@@ -31,6 +31,11 @@ export function useQueue<T[]>(initialValue: T[] =[]): {
 
 #### Arguments
 
-- _**initialValue**_ _`T[]`_ - The initial value of the queue state. Defaults to an empty array.
+- **initialValue** _`T[]`_ - The initial value of the queue state. Defaults to an empty array.
 
 #### Return
+
+- **first** _`T`_ - The value of the first item in the queue.
+- **last** _`T`_ - The value of the last item in the queue.
+- **add** _`(item: T) => void`_ - Adds a value to the end of the queue.
+- **remove** _`() => T`_ - Removed and returns the value from the end of the queue
--- a/src/useQueue/__docs__/story.mdx
+++ b/src/useQueue/__docs__/story.mdx
@@ -39,3 +39,4 @@ export function useQueue<T[]>(initialValue: T[] =[]): {
 - **last** _`T`_ - The last item in the queue.
 - **add** _`(item: T) => void`_ - Adds an item to the end of the queue.
 - **remove** _`() => T`_ - Removes and returns the head of the queue.
+- **size** _`number`_ - The current size of the queue.
--- a/src/useQueue/__docs__/story.mdx
+++ b/src/useQueue/__docs__/story.mdx
@@ -2,9 +2,11 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
 import { Example } from './example.stories';
 import { ImportPath } from '../../storybookUtil/ImportPath';
 
-<Meta title="New Hook/useQueue" component={Example} />
+<Meta title="State/useQueue" component={Example} />
 
 # useQueue
+
+A react state hook implements simple first-in first-out (FIFO) queue.
 
 #### Example
 
@@ -15,7 +17,12 @@ import { ImportPath } from '../../storybookUtil/ImportPath';
 ## Reference
 
 ```ts
-
+export function useQueue<T[]>(initialValue: T[] =[]): {
+  add: (item: T) => void;
+  remove: () => T;
+  first: T;
+  size: number;
+}
 ```
 
 #### Importing
@@ -23,5 +30,7 @@ import { ImportPath } from '../../storybookUtil/ImportPath';
 <ImportPath />
 
 #### Arguments
+
+- _**initialValue**_ _`T[]`_ - The initial value of the queue state. Defaults to an empty array.
 
 #### Return
--- a/src/useQueue/__docs__/story.mdx
+++ b/src/useQueue/__docs__/story.mdx
@@ -6,7 +6,7 @@ import { ImportPath } from '../../__docs__/ImportPath';
 
 # useQueue
 
-A state hook in the form of a first-in first-out (FIFO) queue.
+A state hook in the form of a queue. Can be either first-in-first-out (FIFO) or last-in-last-out (LILO).
 
 #### Example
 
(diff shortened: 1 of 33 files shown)


src/useQueue/__docs__/story.mdx
```diff
@@ -0,0 +1,41 @@
...
+#### Arguments
+
+- **initialValue** _`T[]`_ - The initial value of the queue state. Defaults to an empty array.
+
+#### Return
+
+- **first** _`T`_ - The value of the first item in the queue.
+- **last** _`T`_ - The value of the last item in the queue.
+- **add** _`(item: T) => void`_ - Adds a value to the end of the queue.
+- **remove** _`() => T`_ - Removed and returns the value from the end of the queue
```
ArttuOll, line 41: This description seems to be wrong: the value is removed from the beginning, not the end, so this should be: "Removes and returns the head of the queue".

src/useQueue/__tests__/dom.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(result.error).toBeUndefined();
+  });
+
+  it('should accept an initial value', () => {
+    const { result } = renderHook(() => useQueue([0, 1, 2, 3]));
+    expect(result.current.first).toBe(0);
+  });
+});
+
+describe('remove', () => {
```
ArttuOll, line 20: One describe block probably would suffice for the whole test file since there's only one test for each of the functions.

src/useQueue/useQueue.ts
```diff
@@ -0,0 +1,52 @@
...
+      return state[0];
+    },
+    get last() {
+      return state[state.length - 1];
+    },
+    get size() {
+      return state.length;
+    },
+  };
+};
```
ArttuOll, line 52: The returned functions here are not referentially stable. Their reference will change between renders, which will cause problems, for example, an infinite loop if used inside of `useEffect`.

You can use `useMemo` to achieve referential stability. You can also take a look at some of our other hooks to see how they've achieved this. For example, see [useAsync](https://github.com/react-hookz/web/blob/master/src/useAsync/useAsync.ts).

I pushed a test to this PR, which tests for referential equality of the returned functions.

ArttuOll (changes requested):
Good job all in all. Once the review comments are addressed this could be merged.

However, I started to wonder: this hook is basically an abstraction on top of `useList`. This hook could be implemented with it in a pretty straightforward way. @xobotyi What do you think, should we add hooks which are abstractions of other hooks? In the past, quite a few hooks have been dropped because they could be implemented directly with something like `useMediatedState`, for example.

src/useQueue/__docs__/story.mdx
jpwallace22: Oops 😅

src/useQueue/__tests__/dom.ts
ArttuOll: This really is not a big thing. It's just a matter of grouping the tests logically (many tools also use the groupings to display the test results). To me it seemed odd that there's only one test per group and I haven't seen that anywhere else in the codebase.

src/useQueue/useQueue.ts
jpwallace22: Good call out! I didn't even think of this. Sorry I didn't check the "allow edits from maintainers" box. It's checked now. I went ahead and added this test as well as a Memoization of the return.

jpwallace22:
Thanks for the review @ArttuOll. I really hope you'll decide to merge it in. This is my first contribution to open source. I used 'react-use' quite a bit until the SSR bit started biting me in the butt. So I thought "If I want good hooks, might as well start helping out" haha.

ArttuOll:
I've thought about this and I can't help but think that if this hook is added, it should be implemented with `useList`, because it is just a limited version of it. That way, if we fix bugs in `useList`, the bugs would also get fixed in this hook. This is one of the reasons why it's preferrable to compose previously implemented hooks to implement new ones.

Could you do this @jpwallace22 ?

ArttuOll (approved):
The returned methods required a bit more work in order to be referentially stable, but that's my fault, because the test I gave you was actually faulty! When I wrote the test, I did not take into account that it was actually the invokation of the methods which caused them to get new references instead of just plain rerendering of the component using the hook.

I fixed the test and adjusted the implementation to pass the test. I also edited the doc comments of the hook to be in line with the story of the hook.

jpwallace22:
> The returned methods required a bit more work in order to be referentially stable, but that's my fault, because the test I gave you was actually faulty! When I wrote the test, I did not take into account that it was actually the invokation of the methods which caused them to get new references instead of just plain rerendering of the component using the hook.
> 

Thanks for taking care of that and explaining it to me as well. Like I said, I'm new to testing and reasonable unfamiliar with referential stability. Though now reading up a lot on it. Everyday is a learning day!

ArttuOll:
@xobotyi I asked you this once already, but I think you forgot to answer: are other maintainers allowed to merge PRs or do you want everything to go through you?

I know that maintainer's are allowed to merge their own PRs after they've been reviewed by some other maintainer, but that's a different case.

jpwallace22:
@ArttuOll @xobotyi I'd love to get this merged in. Are there any actionable items I can help with to get it merged?

xobotyi:
@ArttuOll it depends on hook tbh) for now let's pass everything through me.

xobotyi:
SO!
1) as this hook does not implement any queue, nor has switch to change the type of queue - it has to be named `useFifoQueue` or has to be polimorphic, meaning it has to be switchable between fifo\lifo
2) i would expect list itself to be exposed, to be able to peek the items at any position without need to remove them. I'd expect overall api to be like that:
    ```ts
    interface UseQueueResponse<T> {
      // internal array itself
      items: T[];
      add: (item: T) => void;
      remove: () => T;
      first: T;
      last: T;
      size: number;
    }
    ```
    Yes it kind of compromises abstraction and allows arbitrary changes to the queue, but it is better as allows weird implementation not covered by base queue algorithm.

jpwallace22:
@xobotyi Thanks for the feedback! Good callouts all around. 
I implemented the feedback by adding a fifo/lilo option. Seemed the better approach than just remaining it and limiting the hook. I also added the `items` property to the return. 
One thing to note, I DID make it LILO and not LIFO as you specified, but I'm 99% sure that was a typo as that wouldn't be a queue. 😆
Let me know if you need anything else 🙌🏽

ArttuOll:
Looks good to me. I updated the docs a little.

> Yes it kind of compromises abstraction and allows arbitrary changes to the queue, but it is better as allows weird implementation not covered by base queue algorithm.

If we want to prevent arbitrary changes to the queue (which btw don't cause a rerender, because `useList` is used under the hood), we could substitute `items` with a `peek: (n: number) => T[]` method.

xobotyi:
Despite of getting the point of LILO (the other way of naming FIFO) i'm hearing this obbreviation for the first time. 
I'd rather use common names for things that are used in CS - therefore it should be named FIFO

xobotyi:
Reading your comment again - **`LIFO != LILO`** LIFO is a 'stack' quque, when items pushed to the back and removed from the back, while FIFO is 'line' queue where items pushed to the back and removed from the front.

ArttuOll:
> Reading your comment again - **`LIFO != LILO`** LIFO is a 'stack' quque, when items pushed to the back and removed from the back, while FIFO is 'line' queue where items pushed to the back and removed from the front. Both are queues.

I think @jpwallace22  was concerned that LIFO queue is commonly considered to be a stack and thus not suitable for this hook. However, LIFO queue seems to be a synonym for stack (makes sense, of course, when you think about it. I also hadn't ever thought of it this way) so it could be included here.

jpwallace22:
@ArttuOll @xobotyi At this point, I'm not too sure what you all want done to this. There seems to be a lot of back and forth. 

FIFO != LILO. They are both queues, but going in different directions. I have never heard of a stack being called a queue, so I assumed `lifo` was a type-o. I don't agree with adding the ability `lifo` for the `useQueue`. That would go against standard CS language. Would you prefer I remove all the 'lilo' logic and change the name to useFifoQueue?

xobotyi:
@jpwallace22 welp here comes language differences =) since both lifo and fifo are queuing containers there comes confusion between CS and natural language🙃
ill tidy things up and merge in couple of hours i hope

PR #1049, opened 2022-12-18
Add `useDocumentVisibility`

Myzel394:
## What new hook does?

`useWindowVisibility` allows you to track whether the window is currently visible or not. I went with `useWindowVisibility` instead of `useVisibility` because this could lead to confusion as some people could think this hook would track the visibility of single React components for example. This hooks uses the `document.visibilityState` API.

## Checklist

- [ ] Does the code have comments in hard-to-understand areas? **(the code is fairly simple)**
  - _link issue here_

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,8 +195,8 @@ Coming from `react-use`? Check out our
   - [**`useScreenOrientation`**](https://react-hookz.github.io/web/?path=/docs/sensor-usescreenorientation--example)
     — Checks if the screen is in `portrait` or `landscape` orientation and automatically re-renders on
     orientation change.
-  - [**`useWindowVisibility`**](https://react-hookz.github.io/web/?path=/docs/sensor-usewindowvisibility--example)
-    — Check if the window is currently visible.
+  - [**`useDocumentVisibility`**](https://react-hookz.github.io/web/?path=/docs/sensor-usedocumentvisibility--example)
+    — Tracks document visibility state.
 
 - #### Dom
 
--- a/README.md
+++ b/README.md
@@ -195,6 +195,8 @@ Coming from `react-use`? Check out our
   - [**`useScreenOrientation`**](https://react-hookz.github.io/web/?path=/docs/sensor-usescreenorientation--example)
     — Checks if the screen is in `portrait` or `landscape` orientation and automatically re-renders on
     orientation change.
+  - [**`useWindowVisibility`**](https://react-hookz.github.io/web/?path=/docs/sensor-usewindowvisibility--example)
+    — Check if the window is currently visible.
 
 - #### Dom
 
--- a/src/useDocumentVisibility/__docs__/example.stories.tsx
+++ b/src/useDocumentVisibility/__docs__/example.stories.tsx
@@ -1,9 +1,9 @@
 import * as React from 'react';
 import { useEffect } from 'react';
-import { useWindowVisibility } from '../..';
+import { useDocumentVisibility } from '../..';
 
 export const Example: React.FC = () => {
-  const isVisible = useWindowVisibility();
+  const isVisible = useDocumentVisibility();
 
   useEffect(() => {
     if (!isVisible) {
--- a/src/useDocumentVisibility/__docs__/example.stories.tsx
+++ b/src/useDocumentVisibility/__docs__/example.stories.tsx
@@ -8,7 +8,7 @@ export const Example: React.FC = () => {
   useEffect(() => {
     if (!isVisible) {
       // eslint-disable-next-line no-alert
-      alert('Window was not visible');
+      alert('Document was not visible');
     }
   }, [isVisible]);
 
--- a/src/useDocumentVisibility/__docs__/story.mdx
+++ b/src/useDocumentVisibility/__docs__/story.mdx
@@ -2,13 +2,11 @@ import { Canvas, Meta, Story } from '@storybook/addon-docs';
 import { Example } from './example.stories';
 import { ImportPath } from '../../__docs__/ImportPath';
 
-<Meta title="Lifecycle/useWindowVisibility" component={Example} />
+<Meta title="Lifecycle/useDocumentVisibility" component={Example} />
 
-# useWindowVisibility
+# useDocumentVisibility
 
 Returns a boolean indicating whether the window is visible or not.
-
-> **_This hook provides stable API, meaning returned methods does not change between renders_**
 
 #### Example
 
@@ -19,7 +17,7 @@ Returns a boolean indicating whether the window is visible or not.
 ## Reference
 
 ```ts
-function useWindowVisibility(ssrValue = true): () => boolean;
+function useDocumentVisibility(initializeWithValue = true): () => boolean;
 ```
 
 #### Importing
@@ -28,7 +26,7 @@ function useWindowVisibility(ssrValue = true): () => boolean;
 
 #### Arguments
 
-- **ssrValue** _`boolean`_ _(default: `true`)_ - Value to return on server-side rendering.
+- **initializeWithValue** _`boolean`_ _(default: `true`)_ - Value to return on server-side rendering.
 
 #### Return
 
--- a/src/useDocumentVisibility/__docs__/story.mdx
+++ b/src/useDocumentVisibility/__docs__/story.mdx
@@ -1,33 +1,31 @@
-import { Canvas, Meta, Story } from '@storybook/addon-docs';
-import { Example } from './example.stories';
-import { ImportPath } from '../../__docs__/ImportPath';
+import {Canvas, Meta, Story} from '@storybook/addon-docs';
+import {ImportPath} from '../../__docs__/ImportPath';
+import {Example} from './example.stories';
 
-<Meta title="Lifecycle/useDocumentVisibility" component={Example} />
+<Meta title="Lifecycle/useDocumentVisibility" component={Example}/>
 
 # useDocumentVisibility
 
-Returns a boolean indicating whether the window is visible or not.
+Returns a boolean indicating whether the document is visible or not.
 
 #### Example
 
 <Canvas>
-  <Story story={Example} inline />
+    <Story story={Example} inline/>
 </Canvas>
 
 ## Reference
 
 ```ts
-function useDocumentVisibility(initializeWithValue = true): () => boolean;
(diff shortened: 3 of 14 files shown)


Myzel394:
While reviewing this PR, I thought maybe `useDocumentVisibility` would fit better, what do you think?

ArttuOll:
Hello and thanks for the PR! Reviewing this might take a while since the Christmas holidays are coming, but for now, let's run the workflows.

src/useWindowVisibility/__docs__/story.mdx
```diff
@@ -0,0 +1,35 @@
...
+function useWindowVisibility(ssrValue = true): () => boolean;
+```
+
+#### Importing
+
+<ImportPath />
+
+#### Arguments
+
+- **ssrValue** _`boolean`_ _(default: `true`)_ - Value to return on server-side rendering.
```
ArttuOll, line 31: Better rename this to `initializeWithValue` to be in line with our other hooks.

Also, the functionality should be as follows:
- If `initializeWithValue` is `true`, return the value of the hook on first render.
- If `initializeWithValue` is `false`, return `undefined` on first render.

Users should set `initializeWithValue` to `false` during SSR.

src/useWindowVisibility/useWindowVisibility.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { useEventListener } from '../useEventListener/useEventListener';
+import { isBrowser } from '../util/const';
+
+/**
+ * Returns a boolean indicating whether the window is visible or not.
+ * @param ssrValue Value to return on server-side rendering.
+ */
+export function useWindowVisibility(ssrValue = true): boolean {
+  const [isVisible, setIsVisible] = useState(document.visibilityState === 'visible');
```
ArttuOll, line 11: This line fails during SSR (as evidenced by the SSR tests failing), because `document` is not available during SSR. You have to check that the hook is being run in a browser before calling browser APIs.

For reference, see the implementation of [useWindowSize](https://github.com/react-hookz/web/blob/master/src/useWindowSize/useWindowSize.ts) (especially lines 34 and 35).

src/useWindowVisibility/useWindowVisibility.ts
```diff
@@ -0,0 +1,24 @@
...
+  const handleVisibilityChange = useCallback(() => {
+    setIsVisible(document.visibilityState === 'visible');
+  }, []);
+
+  useEventListener(document, 'visibilitychange', handleVisibilityChange);
+
+  if (!isBrowser) {
+    return ssrValue;
+  }
+
```
ArttuOll, line 22: This block becomes unnecessary if `isVisible` is initialized properly (see my comment on line 11).

src/useWindowVisibility/__docs__/story.mdx
```diff
@@ -0,0 +1,35 @@
...
+import { Example } from './example.stories';
+import { ImportPath } from '../../__docs__/ImportPath';
+
+<Meta title="Lifecycle/useWindowVisibility" component={Example} />
+
+# useWindowVisibility
+
+Returns a boolean indicating whether the window is visible or not.
+
+> **_This hook provides stable API, meaning returned methods does not change between renders_**
```
ArttuOll, line 11: This hook does not return any methods, so this line should be removed.

ArttuOll (changes requested):
Ok, I could not resist doing a quick review anyway!

ArttuOll:
> While reviewing this PR, I thought maybe `useDocumentVisibility` would fit better, what do you think?

Agreed. `useDocumentVisibility` would be a more descriptive name.

src/useWindowVisibility/__docs__/story.mdx
Myzel394: > 
    If initializeWithValue is true, return the value of the hook on first render.
    If initializeWithValue is false, return undefined on first render.

Why?

xobotyi:
Sorry for depretion of done work, i really thankful for the pr, but such behaviour is doable with two lines of code
```ts
const isVisible
```

xobotyi:
Really sorry for close, missclicked🤦‍♂️

src/useWindowVisibility/__docs__/story.mdx
xobotyi: Because ther is no need to emulate any behaviour since for SSR users we're able to return undefined on first render, avoiding hydration mismatch, and delaying state fetch to evvects phase.

src/useDocumentVisibility/useDocumentVisibility.ts
```diff
@@ -0,0 +1,24 @@
+import { useCallback, useState } from 'react';
+
+import { useEventListener } from '../useEventListener/useEventListener';
+import { isBrowser } from '../util/const';
+
+/**
+ * Returns a boolean indicating whether the window is visible or not.
+ * @param initializeWithValue Value to return on server-side rendering.
+ */
+export function useDocumentVisibility(initializeWithValue = true): boolean {
```
xobotyi, line 10: I propose below implementation
```ts
const isDocumentVisible = () => document.visibilityState === 'visible';

/**
 * Returns a boolean indicating whether `document.visibilityState` equals `visible`.
 * @param initializeWithValue Whether to initialize hook with value on first
 * render or postpone it to effects stage. SSR users should set it to `false`.
 */
export function useDocumentVisibility(initializeWithValue = true): boolean | undefined {
  initializeWithValue = isBrowser && initializeWithValue;
  const [isVisible, setIsVisible] = useState(initializeWithValue ? isDocumentVisible() : undefined);

  useMountEffect(() => {
    if (initializeWithValue) {
      setIsVisible(isDocumentVisible());
    }
  });

  useEventListener(window, 'visibilitychange', () => setIsVisible(isDocumentVisible()));

  return isVisible;
}
```

xobotyi:
1) It made through promise for the case when developer wants to change behavior.
2) `useRafState` is not safer it is lise usual `useState` but new state is only appliad once in a animation frame (it is written in hook docs).
3) Resize event is called a lot during resize process - most likely we dont need to rerender our app that much and want to resuce amount of rerenders - therefore raf state is used.
    This hook does not require such throttling since visibility change is not able to change that fast.

README.md
```diff
@@ -193,6 +193,8 @@ Coming from `react-use`? Check out our
   - [**`useScreenOrientation`**](https://react-hookz.github.io/web/?path=/docs/sensor-usescreenorientation--example)
     — Checks if the screen is in `portrait` or `landscape` orientation and automatically re-renders on
     orientation change.
+  - [**`useWindowVisibility`**](https://react-hookz.github.io/web/?path=/docs/sensor-usewindowvisibility--example)
```
xobotyi, line 196: You forgot to rename it here.

src/useDocumentVisibility/__docs__/example.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useEffect } from 'react';
+import { useDocumentVisibility } from '../..';
+
+export const Example: React.FC = () => {
+  const isVisible = useDocumentVisibility();
+
+  useEffect(() => {
+    if (!isVisible) {
+      // eslint-disable-next-line no-alert
+      alert('Window was not visible');
```
xobotyi, line 11: Better to address it as `Document` because of hook name.

src/useDocumentVisibility/__docs__/story.mdx
```diff
@@ -0,0 +1,33 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs';
+import { Example } from './example.stories';
+import { ImportPath } from '../../__docs__/ImportPath';
+
+<Meta title="Lifecycle/useDocumentVisibility" component={Example} />
+
+# useDocumentVisibility
+
+Returns a boolean indicating whether the window is visible or not.
```
xobotyi, line 9: same here - not window, document.

ArttuOll:
> Can one of you explain why we use this approach and whether it would be applicable to also use it in useDocumentVisibility?

You don't need to use it here. It's not relevant to this hook. My understanding is that `useRafState` is only used with dynamically changing visual elements, eg. animations and such.

xobotyi (changes requested):

xobotyi:
@ArttuOll 
> My understanding is that useRafState is only used with dynamically changing visual elements, eg. animations and such.

Not only - it also usable for cases when you have high frequency events and dont want to overload browser with re-renders, treat it as throttle that synchronized by browser to achieve 60fps rendering.

src/useDocumentVisibility/useDocumentVisibility.ts
```diff
@@ -0,0 +1,34 @@
...
+    // eslint-disable-next-line unicorn/consistent-function-scoping
+    const handleVisibilityChange = () => setIsVisible(isDocumentVisible());
+
+    document.addEventListener('visibilitychange', handleVisibilityChange, {
+      passive: true,
+    });
+
+    return () => {
+      document.removeEventListener('visibilitychange', handleVisibilityChange);
+    };
```
ArttuOll, line 30: As xobotyi suggested, it would be beneficial to use `useEventListener` here, because it already handles all of this logic. In general, we prefer reusing our own hooks as much as possible, because that way the resulting hooks become more maintainable and improvements in one hook lead to improvements in other hooks as well.

ArttuOll: All of our hooks do.

Also, I meant to only use `useWindowSize` as reference for implementing the setting of the initial value.

Myzel394: Hey @ArttuOll with my last commit this hook now uses `useEventListener` to listen to events. However, I don't know how to make this SSR-friendly, as `document` is not available on the server. I thought about conditionally calling the hook, but this breaks the rules. I think going with my initial idea (manually registering events) would be the best option in this case, what do you think?

xobotyi: Do not reference web-apis during SSR and everything will be fine🙃
i gave you solution here: https://github.com/react-hookz/web/pull/1049#discussion_r1051625771

In case we're not in browser (therefore were in ssr) - `document` not referenced, effects are not invoked during ssr therefore there is no need in guarding there.

PR #1063, opened 2023-01-02
feat(useMemoCache): useMemo with cache

michaltarasiuk:
## Hook description

Like useMemo but with cache based on dependency list.

### Valid use-case for the hook

Avoid expensive callculation with the same dependencies

## Checklist

- [X] Have you read the [contribution guidelines](../../CONTRIBUTING.md)?
- [ ] If you are porting a hook from `react-use`, have you checked #33 and the [migration guide](../../src/__docs__/migrating-from-react-use.story.mdx)
  - _link issue here_
- [X] Have you written tests for the new hook?


ArttuOll:
Please use the provided PR template.

xobotyi:
And from the start about implementation - this implementation has several major issues
1) The cache is not guarded from eating all memory
2) The cache is slow since it is using array as storage and it's complexity of access is `O(n)`
3) You dont provide hatches to alter dependencies comparison

MichalTarasiuk:
Hello, I have to create a hook which will be used by `useMemoCache`. Pull request will be created in days.

MichalTarasiuk:
Duplicated

ArttuOll:
Thanks for the PR! I will have time to review this over the weekend. Someone might also review it before me.

For now, could you expand a little on the valid use-case for this hook? What is the benefit over `useMemo`? Your description doesn't tell much and the docs don't either.

ArttuOll:
> I will have time to review this over the weekend.

I really want to have a discussion on the use-cases of this hook before reviewing this (and before any more code is written) to make sure that no unnecessary work is done by anyone. At the moment, it is not clear to me what the use-case for this hook is.

xobotyi:
The usecase is pretty simple - there are no unncessary invocations of memo factory in case provided deps are the same.

4ex. in case we had as dep value 2 then 3 then 2 again - this hook will fast-return result of invocation when dep was 2 without calling it again

so it is like good-old memoization

ArttuOll:
> The usecase is pretty simple - there are no unncessary invocations of memo factory in case provided deps are the same.
> 
> 4ex. in case we had as dep value 2 then 3 then 2 again - this hook will fast-return result of invocation when dep was 2 without calling it again
> 
> so it is like good-old memoization

I see, thanks. So this would be good if one was doing something really heavy in the factory function.

xobotyi:
@MichalTarasiuk you havent resolved issues - you just accepted all your's - there were changes to import paths and output bundle - your PR now will roll back all changes we've made🙃 Reread contribution guide please and appropriately rebase your changes.

MichalTarasiuk:
Hello, code does not have 100% cover because value of `objectIs` variable is being assigned before we import and it is not possible to mock.

ArttuOll:
@MichalTarasiuk Could you notify us when this is ready for review or is it already?

MichalTarasiuk:
> @MichalTarasiuk Could you notify us when this is ready for review or is it already?

Hello @xobotyi @ArttuOll, ready for code review.

src/useMemoCache/__docs__/example.stories.tsx
```diff
@@ -0,0 +1,57 @@
...
+      </section>
+      <section>
+        <h1>Memo with cache</h1>
+        <p>summary: {memoCache}</p>
+        <p>calls: {memoCacheSpy.current}</p>
+      </section>
+      <button onClick={toggleDependencyList}>toggle dependency list</button>
+    </div>
+  );
+};
```
ArttuOll, line 57: The example is good, but I think the `summary` value just being toggled between 4 and 6 is really confusing. I think the example needs to do something a bit more realistic to really illustrate the functionality of the hook. Maybe you could do something like let the user input a number and then count all Fibonacci numbers until that point while displaying the number of calls made to the factory functions below.

Also, make sure that the example is understandable by using the example program and reading the example code. Storybook doesn't show all code that you have in the `example.stories.tsx` file. For example, now `getSummary` and `dependencyListMapper` are not visible in the example code, which prevented me from understanding the example when reading through Storybook.

Lastly, `getSummary` should be renamed to `sum`, if you intend to keep it.

src/useMemoCache/__docs__/story.mdx
```diff
@@ -0,0 +1,28 @@
+import { Canvas, Meta, Story } from '@storybook/addon-docs/blocks';
+import { Example } from './example.stories';
+import { ImportPath } from '../../__docs__/ImportPath';
+
+<Meta title="Miscellaneous/useMemoCache" component={Example} />
+
+# useMemoCache
+
```
ArttuOll, line 8: Hook description missing here. You have it in other places and it's good.

src/useMemoCache/__docs__/story.mdx
```diff
@@ -0,0 +1,28 @@
...
+```
+
+#### Importing
+
+<ImportPath />
+
+#### Arguments
+
+- **factory** _`() => unknown`_ - useMemo factory function.
+- **deps** _`DependencyList`_ - useMemo dependency list.
```
ArttuOll, line 28: The function reference and the arguments description are missing the `customAreHookInputsEqual` argument.

src/useMemoCache/__tests__/dom.ts
```diff
@@ -0,0 +1,209 @@
...
+    expect(result.current).toEqual([1, 2]);
+    expect(spy).toHaveBeenCalledTimes(1);
+
+    rerender({ dependencyList: [{ a: 2, b: 3 }] });
+
+    expect(result.current).toEqual([2, 3]);
+    expect(spy).toHaveBeenCalledTimes(2);
+  });
+
+  it('should handle unstable refference of `areHookInputsEqual`', () => {
```
ArttuOll, line 114: __reference__

src/useMemoCache/index.ts
```diff
@@ -0,0 +1,98 @@
...
+    const state = factory();
+
+    memoCache.set(deps, state);
+
+    return state;
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, deps);
+
+  return memo;
+};
```
ArttuOll, line 98: I don't see anywhere the part which would prevent the cache from becoming infinitely large. As requested earlier, you should add such a limiter, which would allow, say, 64 entries, and write tests for it.

ArttuOll (changes requested):
You've done a thorough job! I noticed some things that need to be fixed and @xobotyi might find some more. Anyhow, I'm sure this will be merged eventually.

MichalTarasiuk:
Hello @ArttuOll @xobotyi, Code review fixes are added, but I not sure how guard from eating all memory should be implemented.

ArttuOll:
> Hello @ArttuOll @xobotyi, Code review fixes are added, but I not sure how guard from eating all memory should be implemented.

Looking at the code quickly in the browser, I think you can just not allow the `cache` (which is a `Map`) to have over 64 entries. You can check the cache size in `set`. Something like:
```typescript
if (!hasCachedItem && cacheHasSpaceLeft) {
...
}
```

MichalTarasiuk:
Hello @ArttuOll @xobotyi, logic of max entries is ready 🎉.

MichalTarasiuk:
Hello @ArttuOll @xobotyi, logic of max entries is ready 🎉.

ArttuOll (changes requested):

src/useMemoCache/index.ts
```diff
@@ -14,7 +14,10 @@ type CachedItem<State> = { state: State; dependencyList: DependencyList };
 export const createMemoCache = <State>(
   customAreHookInputsEqual?: typeof nativeAreHookInputsEqual
 ) => {
-  const cache = new Set<CachedItem<State>>();
+  const MAX_ENTRIES = 64;
+  let indexCounter = 0;
```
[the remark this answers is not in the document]
MichalTarasiuk: Hello, `indexCounter` is used to determine which element should be removed when threshold is hit

MichalTarasiuk:
Changed to use `Array` instead of `Map`. The overall number of elements stays the same, but indexes now range from 0 - 63.

src/useMemoCache/index.ts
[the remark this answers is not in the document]
ArttuOll: Oh, I see. Now I wonder if it would be better to just stop accepting new items to the cache after the maximum size is reached since the cache items are basically equally "important" so there is no reason to be deleting one item instead of another when the cache is filled.

ArttuOll:
> Changed to use `Array` instead of `Map`. The overall number of elements stays the same, but indexes now range from 0 - 63.

This brings the time complexity of accessing the cache back to O(n), which is not ideal.

MichalTarasiuk:
Previously i haven't used advantages that `Map` gives. Search complexity is O(n^2), but i think it doesn't make sense to optmize since it can hold only 64 items (insertion is O(1)).

MichalTarasiuk:
Hello, please let me know if there is any way of use `Map`. I don't have any idea. 

In previous approach I didn't use advantages of `Map`, because when function `get` of `createCache` instance was invoked the `Map` was converted to `Array` in order to iterate over values.

ArttuOll:
> Hello @ArttuOll @xobotyi, please let me know if there is any way of use `Map`. I don't have any idea.
> 
> In previous approach I didn't use advantages of `Map`, because when function `get` of `createCache` instance was invoked the `Map` was converted to `Array` in order to iterate over values.

At this point, we're just waiting for what @xobotyi has to say. He's been busy lately, so let's be patient!

PR #1074, opened 2023-01-08
`useResponsiveText`

jpwallace22:
## Hook description

- Describe the new hook. What does it do?
`useResponsiveText` returns a ref that you can use to control the text size of whatever element to to be a percentage of its parent. It can be used for times when you need exact control over the responsiveness of text at every size. 

### Valid use-case for the hook

- How would the hook be used in a real application?
You can see it in this [mock-up](https://octomlmarketingsitemain97959-oc214video.gtsb.io/dummy-video/). The text inside the video component uses it so its always proportionate to the video at ANY size, not just breakpoints. 

## Checklist

- [x] Have you read the [contribution guidelines](../../CONTRIBUTING.md)?
- [x] If you are porting a hook from `react-use`, have you checked #33 and the [migration guide](../../src/__docs__/migrating-from-react-use.story.mdx)
  - _link issue here_
- [x] Have you written tests for the new hook?
  - **note** the test coverage is kind of low because I didn't want to add a library to be able to render components in jest. Making it difficult to test the ref'd element. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,8 @@ Coming from `react-use`? Check out our
     — Subscribes an event listener to a target element.
   - [**`useKeyboardEvent`**](https://react-hookz.github.io/web/?path=/docs/dom-usekeyboardevent--example)
     — Invokes a callback when a keyboard event occurs on the chosen target.
+  - [**`useResponsiveText`**](https://react-hookz.github.io/web/?path=/docs/dom-useresponsivetext--example)
+    — Style referenced element to have a `font-size` thats a percentage of its parent element.
   - [**`useWindowSize`**](https://react-hookz.github.io/web/?path=/docs/dom-usewindowsize--example)
     — Tracks the inner dimensions of the browser window.
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -78,3 +78,5 @@ export { resolveHookState } from './util/resolveHookState';
 
 // Types
 export * from './types';
+
+export * from './useResponsiveText';
--- a/src/useResponsiveText/__docs__/example.stories.tsx
+++ b/src/useResponsiveText/__docs__/example.stories.tsx
@@ -1,6 +1,13 @@
 import * as React from 'react';
-// import { useResponsiveText } from '../..';
+import { useResponsiveText } from '../..';
 
 export const Example: React.FC = () => {
-  return null;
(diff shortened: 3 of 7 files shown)


ArttuOll:
Hello and thanks for the PR!

Before reviewing the code, I would like to ask: what's the benefit of doing this in Javascript (in the form of a React hook) instead of just CSS?

jpwallace22:
> Before reviewing the code, I would like to ask: what's the benefit of doing this in Javascript (in the form of a React hook) instead of just CSS?

@ArttuOll As far as I know, there isn't a way to fit text to be proportionate to a container with CSS. If the container is the `body` you can use `vw` but other than that, you need to use JS.  Unless I am missing something there. And if I am, I would love to know what. Haha. 

Also, as I noted above, I think the test coverage would need another dependency added for testing.

ArttuOll:
You're right, there is no way to do that with CSS, but I'm not sure if such capability is commonly needed. There are many ways to implement responsive text in CSS already (media queries, viewport units, clamp). I'm sure there are situtations where you would need to set the font size strictly based on the width of the parent, but I imagine those would be quite rare (if there is a common situation like this, please let me know).

Personally, I don't think this hook fits this library. What do you think @xobotyi?

jpwallace22:
@ArttuOll I had to use this twice now so I figured it might be worth it to add it here. If @xobotyi OK's it, I'll clean it up and get some better tests. But other than that, no worries at all.

jpwallace22:
@ArttuOll Just bumping before the weekend since I would have time to write the tests this weekend if there was a reasonable use case for this hook.

ArttuOll:
> @ArttuOll Just bumping before the weekend since I would have time to write the tests this weekend if there was a reasonable use case for this hook. 

This one is only waiting for @xobotyi's opinion. He's been busy lately, so let's be patient.

xobotyi:
Hey.
So.. I am really uncomfortable to decline already done job, but here we are🙃

I don't really sure that this hook should be merged because of several:
1. It's usecase really-really rare. 
2. It depends on width only, but some may want it to be height-relative.
3. Such hooks (appearance-specific) should be somewhere in ui-kit, not in "general-purpose" hooks collection.

jpwallace22:
@xobotyi don't feel bad at all. Minus the tests it was mainly just copy and pasting from a project repo.

PR #1088, opened 2023-01-15
Improve docs of the lifecycle category (1/2)

ArttuOll:
## What is the problem?

The documentation contains typos and unclear and even outdated parts.

Go through the documentation and fix typos and unclear parts.

This PR addresses the first half of the lifecycle category (only the first half to avoid this PR from getting too big).

### DIFF
--- a/src/useConditionalEffect/__docs__/example.stories.tsx
+++ b/src/useConditionalEffect/__docs__/example.stories.tsx
@@ -1,54 +1,37 @@
 import * as React from 'react';
 import { useState } from 'react';
-import { useConditionalEffect, useToggle, useUpdateEffect } from '../..';
+import { useConditionalEffect, useUpdateEffect } from '../..';
 
 export const Example: React.FC = () => {
-  const [isToggled, toggle] = useToggle(false);
+  const [state1, setState1] = useState(2);
+  const [state2, setState2] = useState(2);
 
-  // eslint-disable-next-line react/no-unstable-nested-components
-  const ToggledComponent: React.FC = () => {
-    const [state1, setState1] = useState(2);
-    const [state2, setState2] = useState(2);
-
-    useConditionalEffect(
-      () => {
-        // eslint-disable-next-line no-alert
-        alert('COUNTERS VALUES ARE EVEN');
-      },
-      [state1, state2],
-      [state1, state2],
-      (conditions) => conditions.every((i) => i && i % 2 === 0),
-      useUpdateEffect
-    );
-
-    return (
-      <div>
-        <div>Alert will be displayed when both counters values are even</div>
-        <div>Effect also invoked on initial mount</div>
-        <button
-          onClick={() => {
-            setState1((i) => i + 1);
-          }}>
-          increment counter 1 [{state1}]
-        </button>{' '}
-        <button
-          onClick={() => {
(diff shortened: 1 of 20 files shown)


src/useConditionalEffect/__docs__/example.stories.tsx
```diff
@@ -1,54 +1,37 @@
...
+      </button>{' '}
+      <button
+        onClick={() => {
+          setState2((i) => i + 1);
+        }}>
+        Increment counter 2 [{state2}]
+      </button>
     </div>
   );
 };
```
ArttuOll, line 37: When reading throught this example and trying to improve the language, I realised that toggled example component here was really confusing and it did not even work as intended, since no alert was shown when the example component was mounted. Thus, I decided to remove it.

src/useCustomCompareEffect/__docs__/story.mdx
```diff
@@ -40,13 +40,12 @@ export function useCustomCompareEffect<
...
-  effect will be triggered when deps change AND conditions satisfy predicate.
-- **comparator** _`DependenciesComparator<Deps> `_ - Function that compares two dependencies
-  arrays, and returns true in case they're equal.
-- **effectHook** _`EffectHook<Callback, Deps, HookRestArgs>`_ - Effect hook that will be used to
-  run callback. Must comply `useEffect` signature, meaning that callback should be placed as first
-  argument and dependencies list as second.
-- **...effectHookRestArgs** _`HookRestArgs`_ - Extra arguments that passed to effectHook. Meaning
-  the arguments that are after 2nd argument.
+- **callback** _`EffectCallback`_ -  Function that will be passed to the underlying effect hook.
+- **deps** _`DependencyList`_ - Dependency list like the one passed to `useEffect`.
```
JoeDuncko, line 44: Is it okay to drop the `If not undefined - effect will be triggered when deps change AND conditions satisfy predicate.` bit?

src/useIntervalEffect/__docs__/story.mdx
```diff
@@ -1,17 +1,17 @@
...
 # useInterval
 
-Like `setInterval` but in form of react hook.
+Like `setInterval` but in the form of a React hook.
 
-- Auto-cancel interval on component unmount.
-- Reset interval on delay change.
-- Callback change dont cause interval reset.
-- Ability to disable interval.
+- Automatically cancel the interval on component unmount.
```
JoeDuncko, line 11: Maybe "Automatically cancel**s** the interval on component unmount."

JoeDuncko (approved):
Thanks so much for doing this. Had a couple comments, but approving because I think it's basically good to go!

src/useCustomCompareEffect/__docs__/story.mdx
ArttuOll: I deleted it out of routine. In this case I should probably add it back since it is worth highlighting here that merely the dependencies chaning will not trigger the effect.

ArttuOll: I think that part is just copy-pasta from `useConditionalEffect`. There is no predicates to be satisfied in `useCustomCompareEffect`. There is the custom comparator which that bit may be referring to, but I don't think that bit is necessary anyway.

PR #1096, opened 2023-01-23
Deprecate useSafeState

ArttuOll:

`useSafeState` is offered as a solution to situations where asynchronous functions attempt to update a component's state when the component has already unmounted, resulting in the famous `Can't perform a React state update on an unmounted component` error.

`useSafeSate` is like `useState`, but it does an `isMounted` check before allowing the state to be set. This, [according to the React team](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html), is an anti-pattern. A more fresh take on the topic, confirmed by Dan Abramov, can be found [here](https://medium.com/doctolib/react-stop-checking-if-your-component-is-mounted-3bb2568a4934). I won't go in to the details of why `isMounted` checks are an anti-pattern, because I don't want to repeat the resources linked above.

Additionally, the `Can't perform a React state update on an unmounted component` error [has been removed in React 18](https://github.com/reactwg/react-18/discussions/82), because it was considered to be misleading and caused people to invent workarounds, which the React team considered harmful (the example of a harmful workaround in the linked discussion is actually almost exactly `useIsMounted`).

For the above reasons, I think `useSafeState` has no job anymore. Lastly, it should also be considered, whether or not `useIsMounted` should be removed also.

This issue was also discussed a while back in #959 .

`useState` should be used instead of `useSafeState`. Asynchronous functions should be cancelled if the component unmounts or they should be ignored, if cancellation is not an option. From the author of the discussion linked above:
> ... We’re removing the warning because people work around it with isMounted which defeats the purpose. So we might as well not warn. But the proper solution is cancellation. Or ignoring if you’re lazy about cancelling.

1. Replace all internal usages of `useSafeState` with `useState`.
2. Remove `useSafeState`.
3. Remove any mentions of `useSafeState` from the documentation and examples.

## Checklist

  - _link issue here_


xobotyi (approved):
Thanks for thorough explanation.

We will let it to marinate a bit more to avoid overwhelming with breaking changes (supposedly we should have a new rule - one breaking change in a month 😅)

ArttuOll:
What do you think about also removing `useIsMounted`? I think the same reasoning applies to it as well. I could add it to this PR.

xobotyi:
That one still have valid usecases, when dealing with external queues, high frequency events etc.

xobotyi:
yup

PR #1098, opened 2023-01-24
Make use*StorageValue take generic as input to stringify function

akd-io:
`use*StorageValue`'s `stringify` function currently takes an input of type `unknown`.

It would be nicer if the `stringify` function took an input of the generic type `Type`, already present on the `UseStorageValueOptions` interface.

It simply changes `unkown` to `Type`.

I am not 100% sure that this won't introduce problems, and as such, this is more of a PR to hear if you guys agree and think I should pursue and create a proper PR, or if there are problems with doing this, and a simple explanation as to why it isn't implemented.

## Checklist

- [x] There is no existing issue for this PR.
- [ ] Have the files been linted and formatted?
- [ ] Have you run the tests locally to confirm they pass?


ArttuOll:
Hello and thanks for the PR!

Did this cause you any trouble or is this just a thing you noticed?

I think changing the type to `T` would make sense. I can't see why the type of the value to be stringified would be different from `defaultValue` and the return type of `parse`. That said, I would have to play with the code to be certain, but I'm currently traveling so that will have to wait.

Running the CI to see if anything breaks.

akd-io:
Thanks for commenting @ArttuOll, I'll take a deeper look once I get time.

I also just noticed, that the return type includes `undefined` even if a `defaultValue` is set in the options. I'll see if I can tighten that type as well.

akd-io:
@ArttuOll Pulled the code today and I think I got the PR ready.

- I had forgotten to change the type in the implementation of `stringify`, so did that in https://github.com/react-hookz/web/pull/1098/commits/f35cb1659937284328f0d73d81ebcd98c4aabdd3
- `lint:fix` knew exactly what I was doing, and removed a now unnecessary type assertion in https://github.com/react-hookz/web/pull/1098/commits/e9aaa1acd677068e7b5f4f9aa37f7464b90f886c
- Fixed an incorrect addition of `undefined` to the type `UseStorageValueValue` in case `Initialize` (the type of the `initializeWithValue` option) isn't set. It should not include `undefined` in this case, because its default value is `true`.

Let me know if I am mistaken about anything, or if you have questions.

I was not able to resolve the correct typing of `UseStorageValueValue` though. Let me explain what I have understood, and why I don't think I can provide a solution.

-----------------

TL;DR: TypeScript doesn't support partial inference of type parameters. And I'll leave this problem behind. Don't read if you don't care, it's not important.

At the moment, `useStorageValue` is defined like so:

```
export function useStorageValue<
  Type,
  Default extends Type = Type,
  Initialize extends boolean | undefined = boolean | undefined // <- Notice default type is = boolean | undefined
>(
  storage: Storage,
  key: string,
  options?: UseStorageValueOptions<Type, Initialize>
): UseStorageValueResult<Type, Default, Initialize> {
  //...
}
```

If you look at this code,

```
const {value} = useStorageValue<number[]>(storage, 'foo', {
  defaultValue: [],
  initializeWithValue: true,
})
// type of value = number[] | undefined
```

You will notice that the type of value is `number[] | undefined` even though we set `initializeWithValue` to `true`. This is because we've only specified a type for the first generic type, `Type` (equal to `number[]`), and not the others, `Default` and `Initialize`. Because a generic type has been specified, TypeScript will not use inference but the default types for the rest of these generics, resulting in `Initialize` being equal to `boolean | undefined` as it doesn't infer `true` from the specified option. TypeScript is all-or-nothing with respect to inference like this.

Now let's try to change `Initialize` of `useStorageValue` to default to `true` to solve this problem:
```
export function useStorageValue<
  Type,
  Default extends Type = Type,
  Initialize extends boolean | undefined = true // <- Notice default type is now = true
>(
  storage: Storage,
  key: string,
  options?: UseStorageValueOptions<Type, Initialize>
): UseStorageValueResult<Type, Default, Initialize> {
  //...
}
```

This fixes the type of `value`, which is now correctly `number[]`. But if we try to set `initializeWithValue` to `false`... You guessed it... We get a type error:
```
useStorageValue<number[]>(storage, 'foo', {
  defaultValue: [],
  initializeWithValue: false, // Error: Type 'false' is not assignable to type 'true'. ts(2322)
})
```

Again, because one generic type has been specified, all inference is turned off, and the type of `initializeWithValue` it now literally `true` even though we input `false`.

This problem, that we can't partially infer type parameters, has been discussed at length here: https://github.com/microsoft/TypeScript/issues/26242

At the moment you are best left to specify all or none of the generic types. I'll personally be specifying no generics and using the `defaultValue` option to let TS infer the value type from it going forward.

I will leave the value typing issue behind. I am not enough of a TypeScript wizard to implement a workaround, and I am not sure you'd want it in your source anyway.

ArttuOll:
Thanks for the thorough explanation! This is definitely not the first time here when we have trouble with TypeScript. We'll probably have to leave the typing issue behind, like you said.

I'll review this properly over the weekend. For now, let's run the workflows.

ArttuOll (approved):
Thanks for the PR! I like what I'm seeing here. Any thoughts @xobotyi ? This shouldn't even break anything for users since we are just removing `undefined`s and `unknown`s which force type checking / null checking.

xobotyi (approved):
Have a gut feeling about change in types of initialize with value, not talking that it is out of scope of this pull request.

but seems good therefore approving

xobotyi:
@ArttuOll you may notice that your merge didnt cause the minor bersion bump - this is because you havent added `feat: ` prefix for commit subject during squashing

ArttuOll:
> @ArttuOll you may notice that your merge didnt cause the minor bersion bump - this is because you havent added `feat: ` prefix for commit subject during squashing

Dang it. It has previously always been there so didn't notice that it was missing. Sorry, my bad. I can sort this tomorrow or should we just include this in the next release?

xobotyi:
no need, i've planned to bump version tomorrow anyway

akd-io:
Thanks a lot guys! ❤️

PR #1113, opened 2023-01-28
feat: add useStateRef hook

michaltarasiuk:
## Hook description

One rudimentary way to measure the position or size of a DOM node is to use a [callback ref](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs). React will call that callback whenever the ref gets attached to a different node. 

Note that we pass [] as a dependency array to useCallback. This ensures that our ref callback doesn’t change between the re-renders, and so React won’t call it unnecessarily, so in current example the callback ref will be called only when the component mounts and unmount. As in result we will not fall into infinite re-renders. 

## Checklist

- [X] Have you read the [contribution guidelines](../../CONTRIBUTING.md)?
- [ ] If you are porting a hook from `react-use`, have you checked #33 and the [migration guide](../../src/__docs__/migrating-from-react-use.story.mdx)
  - _link issue here_
- [X] Have you written tests for the new hook?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -37,6 +37,12 @@ module.exports = {
       extends: ['@react-hookz/eslint-config/react', '@react-hookz/eslint-config/jest'],
     },
     {
+      files: ['**/__tests__/**/*.jsx', '**/__tests__/**/*.tsx'],
+      rules: {
+        'jest/expect-expect': 'off',
+      },
+    },
+    {
       files: ['**/__docs__/**', '**/__tests__/**'],
       rules: {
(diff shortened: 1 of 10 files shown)


xobotyi:
So, you've
- added new dependency (not compatible with our testing pipeline
- globally disabled linting rule without any reasoning
- once again ignored PR guidelines

Last one is somehow "forgivable"

But first two are no-go since they should be separate pull requests with reasoning

MichalTarasiuk:
Hi @xobotyi, in days I will create a pull request for two first points.

PR #1190, opened 2023-04-05
Deps/upgrade to storybook 7

ArttuOll:
Storybook 7 has been released and it has a ton of breaking changes. In this branch, I've done the easy parts of the migrations, but the remaining part consists of migrating all of the docs files of all of our hooks into [the new format](https://github.com/storybookjs/storybook/blob/main/MIGRATION.md#70-docs-changes).

Before committing huge amounts time to this, I would like to discuss, should we do this or something else potentially? Moving to Docusaurus has been suggested before. Now might be the time to do that. We could also just stay in Storybook 6 for now.

What does everybody think?


jpwallace22:
I'm not a huge contributor, but I follow along and contribute when I can. However, I have been using Storybook professionally for a while and can say that this is nothing new. Storybook ALWAYS has breaking changes. I would recommend pivoting a new direction, or just keeping the outdated version. Unless V7 does something we really want to take advantage of.

xobotyi:
Well yes - kind of agree with @jpwallace22 - storybook as for the moment have a feeling of most unreliable dependency out there - almos each update breaks something.

I'd try to lean forwad docsaurus or something similar in case it will allow us to preserve project structure (previously docosaurus dictated all the docs to be in specific subfolder to work) since we anyways have to update all the docs - it might be a good point to change the docs engine.

PR #1194, opened 2023-04-10
Upgrade to TypeScript 5

ArttuOll:
`ts-patch` (the successor of `ttypescript`) has released a beta which adds support for TypeScript 5 and asked consumers to test it. In this branch, I've confirmed that this library works with the new version of `ts-patch`.

I've:
- Replaced `ttypescript` with `ts-patch` (since `ttypescript` is not being developed anymore).
- Fixed a type-narrowing-related bug in `useEventListener` that appeared with TypeScript 5.

This can be merged later, when the release version of `ts-patch` is available. I'll also format the commit messages then.


xobotyi:
There is one thing i dont like about `ts-patch` - in some kind `ts-patch` breaks typecript itself - it is not something built around typescript like `ttypescript`, it injects into TS itself.

xobotyi:
Otherwise - we're not in hurry here. TS 5 broken lots of things out there, i wast even able to switch any of projects at work, bcause of incompatibility.

ArttuOll:
> There is one thing i dont like about `ts-patch` - in some kind `ts-patch` breaks typecript itself - it is not something built around typescript like `ttypescript`, it injects into TS itself.

The question I would ask now is: could we get rid of the dependency altogether? Is `ttypescript` really necessary?

xobotyi:
Weeelp, with some tinkering around linting (that will enforce `.js` extension within imports) and TS configs - it is possible to get rid of plugins now. It was not at the moment `ttsc` been introduced to our repos.

xobotyi:
And maybe you're right.
Better move will be to get rid of plugins alltogether. It will require changing all imports and, i think, switch module resolution to `ESNext`

ArttuOll:
> And maybe you're right. Better move will be to get rid of plugins alltogether. It will require changing all imports and, i think, switch module resolution to `ESNext` (or maybe NodeNext)

Sounds good. I'm totally up for committing time to this.

ArttuOll:
I upgraded TypeScript, removed `ttypescript`, replaced `ttsc` with `tsc` in build scripts, changed module resolution to `NodeNext` and everything appears to work now.

xobotyi:
🤔 does it geerate files that have `.js`  imports?
im pretty much surprised that it dont require.js extension  in imports

ArttuOll:
The generated `import`s and the `requires` don't have any extensions.

xobotyi:
Well that was the point of plugin
We have to enforce extensions otherwise esm won't work.

ArttuOll:
Interesting. I guess we'll have to go with `ts-patch` then? `ttypescript` and such won't work with TypeScript 5.

xobotyi:
_maybe_ we have to consider finally switch to `"type": "module"`

xobotyi:
so here's the plan.
1. process current PR's
2. migrate to new hooks testing library (`@testing-library/react-hooks` -> `@testing-library/react`)
3. migrate to type: module
4. migrate from storybook to docosurus (supposedly, not sure yet if docosaurus is viable with our directorues structure)

xobotyi:
Done long time ago, lost this PR.

PR #1327, opened 2023-08-02
chore: usePermission docs

emmgfx:
## What is the problem?

The usePermission docs doesn't provide the argument list, and the reference is outdated

Update the reference and add an argument

## Checklist

  - _link issue here_


xobotyi:
Closing prior coming docs rework

PR #1343, opened 2023-08-14
chore(docs): identify that useLocalStorageValue is sync'd across tabs

develohpanda:
## What is the problem?
While reading through the migrate from react-use docs I incorrectly determined that the new `useLocalStorageValue` hook was _not_ sync'd across tabs, because of this note

image

Explicitly identify in the `useLocalStorageValue()` docs that the value is kept in sync across tabs.

## Checklist

  - _link issue here_

### DIFF
--- a/src/useLocalStorageValue/__docs__/story.mdx
+++ b/src/useLocalStorageValue/__docs__/story.mdx
@@ -11,6 +11,7 @@ Manages a single LocalStorage key.
 - Uses JSON serialisation to handle non-string values.
 - Tracks window's `storage` event.
 - Synchronized between all hooks on the page with the same key.
+- Synchronized across multiple tabs.
 - SSR compatible.
 
 > **_This hook provides a stable API, meaning the returned methods do not change between renders._**


xobotyi:
I believe synchronization been implemented after one of refactors of that hook🤔 Bu i forgot to update migration guide.

develohpanda:
@xobotyi I thought that I had misread the migration guide.

Should this be referring to `useLocalStorageValue()` from this package, or `useLocalStorage()` from react-use?

image

xobotyi:
@develohpanda it shuld describe how it works within react-hookz version. After second read i've done it opposite way from the moment i've wrote the comment 🙃
IMO it would be better to rephraze whole note about sync

develohpanda:
Sounds good, thanks for confirming! I'll update and push to this branch.

xobotyi:
As all docs being stripped from sourcecode prior to docs reimplementation - closing the PR.

But anyways thanks for contribution.

PR #1377, opened 2023-09-18
Migrate `useWindowSize` tests to `@testing-library/react`

xobotyi:
As `@testing-library/react-hooks` library is abandoned and forces users to switch to `@testing-library/react` simple switch of react versions implies adapting most of tests.

The library change is not a drop-in threfore requires manual rewriting all tests accroding to [migration guide]

As `@testing-library/react` do not offer a solution for SSR hooks testing - i made an ssr harness which i'm planning to extract to separte package lately when all tests moved and harness will prove to be viable.

### DIFF
--- a/src/useWindowSize/__tests__/dom.ts
+++ b/src/useWindowSize/__tests__/dom.ts
@@ -1,6 +1,6 @@
-import { act, renderHook } from '@testing-library/react-hooks/dom';
+import { act, renderHook } from '@testing-library/react';
 import { useState } from 'react';
-import { useWindowSize, type WindowSize } from '../..';
+import { useWindowSize } from '../..';
 
 describe('useWindowSize', () => {
 	beforeEach(() => {
@@ -25,8 +25,9 @@ describe('useWindowSize', () => {
 	});
 
 	it('should render', () => {
-		const { result } = renderHook(() => useWindowSize());
-		expect(result.error).toBeUndefined();
+		expect(() => {
+			renderHook(() => useWindowSize());
+		}).not.toThrow();
 	});
 
 	it('should use provided state hook', () => {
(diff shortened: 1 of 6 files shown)


xobotyi:
@JoeDuncko @ArttuOll requesting some help with this one.
I'll go from bottom, from  `useWindowSize` to `useQueue`

src/useWindowSize/__tests__/dom.ts
```diff
@@ -25,34 +25,29 @@ describe('useWindowSize', () => {
...
 
 	it('should use provided state hook', () => {
 		const { result } = renderHook(() => useWindowSize(useState));
 
 		expect(result.current.width).toBe(100);
 		expect(result.current.height).toBe(100);
-		expect(result.all.length).toBe(1);
 
-		triggerResize('width', 200);
+		triggerResize('height', 200);
```
malavshah9, line 39: @xobotyi  Why did we changed it to `hieght` from `width`?

xobotyi: I have no idea, to be fair =) it is not intended

There is still an issue with harness - have to rewrite it

malavshah9:
@xobotyi Wanted to contribute in this migration.
Suggest me a approach by which
1. We can modularly collaborate
2. I can participate in Hacktoberfest

xobotyi:
Main issue is that i dont know how to properly migrate to react 18. ideally we have to migrate first to `@testing-library/react` but it requires writing own hook harness and heavily depends on react version.

So it seems that ill have to swrite harnesses for react 17, then migrate all tests to `@testing-library/react` and then switch to react 18 with rewriting harnesses to support react 18.

PR #1472, opened 2023-12-17
[BREAKING] switch package to module type

xobotyi:
As part of migration to react 18 and docosaurus - i made a decision to switch package to `"type": "module"`.

By itself it leads to following BREAKING CHANGES:
- Hooks are now distrubutes as JS built form TS with target `ESNext` and ESM module resolution. There is no more sense to distribute CJS version as package is ESM-only.
- Consequent of above - no more `esm` and `cjs` subfolders - hooks are importable from `index.js` or its own directory which don't have a prefix anymore, thanks to `exports` directive. All of below examples will lead to same result, choose any on your taste:
  - `import { useFirstMountState } from '@react-hookz/web';`
  - `import { useFirstMountState } from '@react-hookz/web/';`
  - `import { useFirstMountState } from '@react-hookz/web/useFirstMountState/';`
  - `import { useFirstMountState } from '@react-hookz/web/useFirstMountState/index.js';`
  
  Thought is seems not to have subfolder, it is only due to `exports` directive, in real it is 
  `@react-hookz/web/dist/useFirstMountState/index.js`.
- Pakage uses `imports` directive to define path alias `#root` - it stays so even in distributed code, thus, some may be affected in case their bundler configured to somehow handle such alias. Those developer shoud configure import rewriter not to handle `node_modules` or `@react-hookz/web` package exclusively.

Side-effect for current PR - documentation is broken, as storybook 6 is not working within ESM packages and Im planning to switch to docosaurus anyways.


package.json
```diff
@@ -21,28 +21,32 @@
...
+		"#root/*": "./src/*"
+	},
+	"exports": {
+		".": "./dist/index.js",
+		"./*/": "./dist/*/index.js",
+		"./*": "./dist/*"
+	},
+	"engines": {
+		"node": ">=20.0.0"
+	},
```
george1410, line 40: what's the rationale behind requiring node 20+ now? node 18 is still in LTS and will be in widespread use

xobotyi: No particular reason except lockfile differs if being built with 18 and 20 versions.
As i understand it forces users to have node20+ - ill lower requirement to node18 (as it is lowest for ESM packages)

george1410: > As i understand it forces users to have node20+

yes this is the trouble we are having while trying to upgrade

> ill lower requirement to node18

much appreciated, thank you!

xobotyi: `24.0.2` is live

Issue #1548, opened 2024-04-16
`useMeasure` hook elements box-sizing option


m-kolomoyets:
## New Features
Current implementation of the `useMeasure` hook calculates the element bounds based on `content-box` model. It is a little unexpected behaviour when the project and all the layout is based on `border-box` model. Thus, current hook ignores paddings calculating element`s `width` and `height`.

### What is the new or updated feature that you are suggesting?
Suggesting adding the additional hook input parameter which has to define whether calculate `content-box` or `border-box` of the element.

Example of code calculating the `border-box` based content size:

```ts
//...
const [observerHandler] = useRafCallback<UseResizeObserverCallback>((entry) => {
    setMeasures({ width: entry.contentRect.width, height: entry.contentRect.height });
});
// ...
```

And extend this logic into current calculation:
```ts
const [observerHandler] = useRafCallback<UseResizeObserverCallback>((entry) => {
    // For example:
    if (boxSizing === 'border-box') {
         setMeasures({
            height: entry.borderBoxSize[0].blockSize,
            width: entry.borderBoxSize[0].inlineSize,
         });
     } else {
         setMeasures({ width: entry.contentRect.width, height: entry.contentRect.height });
    }
});
```

### Why should this feature be included?
Mostly `border-box` model is used, and `content-box` calculations are unexpected behaviour as expected to calculate the box-sizing model needed. Makes sense to add optional parameter to override the initial box-sizing model of the calculation to prevent unexpected miscalculations and debugging.

xobotyi:
IMO it should be done the other way.
user must be able to provide `matcher` - function that matches captured `ResizeObserverEntry` to our height\width

PR #1603, opened 2025-01-30
feat(useMeasure): support custom measurers in order to support border-box and others

m-kolomoyets:

Current implementation of the `useMeasure` hook calculates the element bounds based on content-box model only. 

It is a little unexpected behavior when the project and all the layout is based on border-box model and trying to calculate measures of the element has paddings. 

Thus, current hook ignores paddings calculating element's width and height.

Giving the possibility to define custom measure matcher model to take the values needed.

By adding the additional optional function argument receives the `ResizeObserverEntry` value and returns the new measures should be applied or `null`. If function returns `null`, the default behavior is applied and content-box measures will be returned.

## Checklist

  - https://github.com/react-hookz/web/issues/1548


src/useMeasure/index.ts
```diff
@@ -12,9 +12,11 @@ export type Measures = {
  * Uses ResizeObserver to track element dimensions and re-render component when they change.
  *
  * @param enabled Whether resize observer is enabled or not.
+ * @param observerEntryMatcher A custom matcher function to define the measures. If returns `null` - default `contentRect` measures are applied
  */
 export function useMeasure<T extends Element>(
 	enabled = true,
+	observerEntryMatcher?: (entry: ResizeObserverEntry) => Measures | null,
```
xobotyi, line 19: Proper solution here would be:
1) to define the type `type Measurer = (entry: ResizeObserverEntry) => Measures`. note there is no option for measure to return `null` as you did, since in case element exists it is measurable (a don't see option for etry to me non-measurable)
2) expose two implementations of measurer - `contentBoxMeasurer` and `borderBoxMeasurer`
3) add parameter that defaults to `contentBoxMeasurer`
4) use that measurer all the time

xobotyi (changes requested):

xobotyi:
Thanks for contribution!

xobotyi (approved):
LGTM

PR #1647, opened 2025-04-11
Fix storage hooks return types

joedelia:
`useSessionStorageValue` with a `defaultValue` and no `initializeWithValue` returns `typeof defaultValue | undefined`.

It should not have `undefined` as a possible return type, as we have a `defaultValue` and `initializeWithValue` defaults to `true`.

We correctly check that `initializeWithValue` is actually `false`.

## Checklist

  - _link issue here_

### DIFF
--- a/src/useStorageValue/index.ts
+++ b/src/useStorageValue/index.ts
@@ -118,7 +118,7 @@ type UseStorageValueValue<
 	Default extends Type = Type,
 	Initialize extends boolean | undefined = boolean | undefined,
 	N = Default extends null | undefined ? null | Type : Type,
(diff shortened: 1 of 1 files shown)


joedelia:
Hey @xobotyi sorry I haven't created an issue first, just went ahead and checked the issue. Let me know if it's cool to merge or if we should fork for our org use! Thank you for maintaining this :)

xobotyi:
Fix by itself is noop, since initial version and yours do end in the same type.
Core issue is that exact type being passed causes all remaining inverence to be disabled and default values being used.

I have idea of the fix, will push it to this pr later.

PR #1663, opened 2025-10-23
fix: remove legacy browser API fallbacks

lennondotw:
## Summary

Removes the remaining code paths that pretend standard browser APIs may be missing. `createQueryEntry` in `useMediaQuery` subscribes with `addEventListener` unconditionally, so the deprecated `removeListener` fallback in `queryUnsubscribe` was asymmetric dead code — and the same "API might not exist" defensiveness lingered in `useNetworkState` and the shared `on`/`off` helpers.

## Changes

- `useMediaQuery`: `queryUnsubscribe` uses `removeEventListener` directly, matching `createQueryEntry` (Safari < 14 fallback removed)
- `useNetworkState`: dropped `mozConnection`/`webkitConnection` — every browser shipping the Network Information API exposes unprefixed `navigator.connection`; Firefox and Safari don't ship it at all
- `util/misc.ts`: `on`/`off` no longer optional-chain `addEventListener`/`removeEventListener` — the `EventTarget` constraint guarantees the methods exist; a non-EventTarget target now throws instead of silently doing nothing
- `usePermission` tests: mocks gained the listener methods a real `PermissionStatus` always has (the silent no-op in `on`/`off` was masking the incomplete mocks)

## Testing

- `yarn lint`
- `yarn test` — full suite, 527 tests across 117 files
- `yarn build`

### DIFF
--- a/src/useMediaQuery/index.ts
+++ b/src/useMediaQuery/index.ts
@@ -47,11 +47,7 @@ const queryUnsubscribe = (query: string, setState: QueryStateSetter): void => {
 		if (dispatchers.size === 0) {
 			queriesMap.delete(query);
 
-			if (mql.removeEventListener) {
-				mql.removeEventListener('change', listener);
-			} else {
-				mql.removeListener(listener);
-			}
+			mql.removeEventListener('change', listener);
 		}
 	}
 };
--- a/src/useNetworkState/index.ts
+++ b/src/useNetworkState/index.ts
@@ -66,12 +66,10 @@ export type UseNetworkState = {
 	type: NetworkInformation['type'] | undefined;
 };
 
-type NavigatorWithConnection = Navigator &
-	Partial<Record<'connection' | 'mozConnection' | 'webkitConnection', NetworkInformation>>;
+type NavigatorWithConnection = Navigator & {connection?: NetworkInformation};
 const navigator = isBrowser ? (globalThis.navigator as NavigatorWithConnection) : undefined;
 
-const conn: NetworkInformation | undefined =
-	navigator && (navigator.connection ?? navigator.mozConnection ?? navigator.webkitConnection);
+const conn: NetworkInformation | undefined = navigator?.connection;
 
 function getConnectionState(previousState?: UseNetworkState): UseNetworkState {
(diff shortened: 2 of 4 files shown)


lennondotw:
Rebased on latest master and updated the fix — the original PR was adding back the `addListener` fallback, but since #1661 intentionally removed it from `createQueryEntry`, the right fix is to also remove the asymmetric `removeListener` fallback from `queryUnsubscribe`. Now both sides consistently use the standard `addEventListener`/`removeEventListener` API.

@xobotyi would you mind taking a look? It's a small cleanup — just removing 4 lines.

lennondotw:
@xobotyi I rebased this PR onto the latest `master` and re-checked whether it is still relevant.

Current `master` still has the asymmetric listener cleanup in `useMediaQuery`: `createQueryEntry` subscribes with `mql.addEventListener(...)` unconditionally, while `queryUnsubscribe` still branches to the deprecated `removeListener` fallback. This PR now only removes that stale fallback, so subscribe/unsubscribe consistently use `addEventListener`/`removeEventListener`.

Validation on the rebased branch:

- `yarn install --immutable`
- `yarn vitest --run src/useMediaQuery/index.dom.test.ts src/useMediaQuery/index.ssr.test.ts` -> 12 tests passed
- `yarn lint src/useMediaQuery/index.ts src/useMediaQuery/index.dom.test.ts src/useMediaQuery/index.ssr.test.ts`
- `yarn build`

The previous Copilot review thread is outdated now, since this PR intentionally removes the legacy fallback rather than preserving it.

xobotyi:
Rebased onto current master and broadened the scope: the same "standard API might be missing" defensiveness existed in `useNetworkState` (vendor-prefixed `connection` fallbacks) and the shared `on`/`off` helpers (optional-chained listener methods), so those are cleaned up here as well. Title and description updated accordingly.

xobotyi:
Thank you for your contribution.
I think i will start a series of reworks focused on non-standard behaviors (which usually is safari related🤦), supporting non-standard would be a task for bundler shimming layer.

Will be released somewhat today, along with deps upgrade

PR #1667, opened 2026-01-11
feat: add useBattery hook

samithahansaka:
## Summary
  - Adds `useBattery` hook that tracks device battery state using the Battery Status API
  - Returns `{ isSupported, fetched, charging, chargingTime, dischargingTime, level }`
  - Subscribes to battery change events
  - SSR-safe with proper cleanup

  ## Test plan
  - [x] DOM tests (8 tests)
  - [x] SSR tests (5 tests)
  - [x] All 540 project tests pass
  - [x] Lint passes
  - [x] Build succeeds

  Partial implementation of #33 (sensor hooks from react-use)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,7 @@ get into your bundle. Direct hook imports should be considered otherwise.
   - [**`useNetworkState`**](./src/useNetworkState/index.ts) — Tracks the state of the browser's network connection.
   - [**`useVibrate`**](./src/useVibrate/index.ts) — Provides vibration feedback using the Vibration API.
   - [**`usePermission`**](./src/usePermission/index.ts) — Tracks the state of a permission.
+  - [**`useBattery`**](./src/useBattery/index.ts) — Tracks the state of the device's battery.
 
 - #### Miscellaneous
   - [**`useSyncedRef`**](./src/useSyncedRef/index.ts) — Like `useRef`, but it returns an immutable ref that contains the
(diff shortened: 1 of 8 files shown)


samithahansaka:
I have rebased this onto the latest `master`. It was out of date and had a conflict, because it changed `vitest.config.ts` and that file is gone now.

What I changed:

- Moved to the new `vp` toolchain. The battery test setup is now registered in `vite.config.ts`.
- Fixed the three review comments from January: cleanup on unmount, handling a failed `getBattery()` call, and the wording of the `fetched` doc.
- Added the hook to the README list. I missed that before.
- Put the four event names in one list, so `on` and `off` cannot get out of sync.
- Squashed everything into one `feat:` commit.

Checks on my machine: `yarn fmt:check`, `yarn lint` and `yarn build` all pass, and 547 tests pass. Coverage for `useBattery` is 94.6% of statements and 77.8% of branches. Two lines are not covered: the early return when the API is missing, and the dev only warning. Both have a comment saying why.

CI has not run here yet. Fork PRs need a maintainer to approve the run.

Two things I would like your opinion on:

- If `getBattery()` fails, the hook returns `{isSupported: true, fetched: false}`. That looks the same as still loading. An `error` field would fix it, but it adds to the public API. I left it out. I can add it if you want.
- `isSupported` is `false` on the server and `true` on the first client render. `useNetworkState` behaves the same way and takes an optional `initialState`. I did not add that here. I can if you want.

`useBattery` is still unchecked in #33. #1394 adds the same hook. I am happy to close this one if you prefer that.

PR #1688, opened 2026-07-22
chore(deps): upgrade eslint to 10 and @ver0/eslint-config to 1.7

xobotyi:
Upgrades `eslint` 9.37.0 → 10.7.0 and `@ver0/eslint-config` 1.3.6 → 1.7.1, with the config migration and lint fallout that entails.

## Packaging/config migration

- `@ver0/eslint-config` 1.7 moved all plugins from bundled dependencies to optional peers — the presets this repo uses (javascript, typescript, react, browser, json, markdown, vitest, prettier) now require their plugins as direct devDependencies (~20 added).
- `buildConfig()` was replaced by per-preset modules; `eslint.config.js` now composes them via `defineConfig()`.

## eslint 10 compat

`eslint-plugin-react` 7.x (latest; peer of xo-react) still calls context APIs removed in eslint 10 — `context.getFilename()` via React version auto-detection and `context.getSourceCode()` in rule factories. Handled by:
- pinning `settings.react.version` (skips the detection path), and
- wrapping the react plugin with `fixupPluginRules` from `@eslint/compat` (targeted — a global `fixupConfigRules` wrap breaks `@eslint/markdown`'s language-plugin rules).

⚠️ **Upstream note for ver0-project/eslint-config:** the react preset requires eslint ^10 but its eslint-plugin-react peer can't run unwrapped on eslint 10 — consider doing the `fixupPluginRules` wrap (or the version-setting pin) inside `configs/react.js` so consumers don't have to.

## New-rule fallout

- `react-hooks/refs` (react-hooks 7 via xo-react 0.29) disabled for src — render-time ref access is the core mechanism of this library (`useSyncedRef`, `useFirstMountState`, `usePrevious`, …).
- `react-hooks/set-state-in-effect` / `react-hooks/use-memo` — per-line disables where the pattern is deliberate (`usePermission`, `useCustomCompareMemo`).
- `@typescript-eslint/strict-void-return` disabled for test files (36 hits, all idiomatic terse arrows in tests).
- `vitest/no-conditional-expect` — real find: 4 SSR test files guarded expects behind `if (result.value !== undefined)`, silently passing when value is undefined; replaced with the asserting `expectResultValue()` helper.
- `require-unicode-regexp` (v flag), `max-nested-callbacks`, prettier reflows, redundant-assertion autofixes.

Supersedes #1684.

Verification: lint 0 problems, build clean, 527/527 tests pass.


xobotyi:
Parked pending the planned migration of linting to oxc — vitest 4 (#1689) and TypeScript 7 (#1691) were restructured off this stack and merged to master directly. If the oxc migration lands first, this PR can be closed; the non-lint fixes it carries (SSR test conditional-expect cleanup, useStorageValue set() typing) may be worth cherry-picking either way.

